PR #7, opened 2017-09-19
Feature/el03 temporary

david8dixon:


src/ACEtk/Tags.hpp
```diff
@@ -0,0 +1,21 @@
+struct El03 {
+
+  struct RadiativeStopping{
+    static constexpr size_t energyGridBegin = 2;
```
david8dixon: Remove Tags.hpp

src/ACEtk/interpretation/EL03/test/el03.test.cpp
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+  }
+}
+
+
+
+
+
+/*
```
david8dixon: remove commented stuff

david8dixon (commented):
Clean-up

PR #11, opened 2018-06-28
Feature/regions

jlconlin:


src/ACEtk/interpretation/Region/src/makeRegions.hpp
```diff
@@ -0,0 +1,12 @@
+// static int makeRegions( ENDFInterpolationParameters eip,
```
apmccartney: file can be deleted

apmccartney:
While I like the Region class you've written, this isn't at all what we discussed this morning

src/ACEtk/interpretation/Region/src/ctor.hpp
```diff
@@ -0,0 +1,14 @@
+Region( int INT, xRange&& x, yRange&& y )
+try:
+  interpolationScheme_( verifyInterpolationScheme( INT ) ),
+  x_( x ), y_( y )
```
apmccartney: You're taking rvalue references to x and y. Don't copy. Move instead.

jlconlin: Yeah, this didn't feel right, but I wasn't quite sure how to do this.

src/ACEtk/interpretation/Region/src/makeRegions.hpp
jlconlin: Placeholder, as discussed.

src/ACEtk/interpretation/ENDFInterpolationParameters.hpp
```diff
@@ -5,6 +5,11 @@ class ENDFInterpolationParameters {
 public:
 #include "ACEtk/interpretation/ENDFInterpolationParameters/src/ctor.hpp"
 
+  auto operator[]( int index ) const {
```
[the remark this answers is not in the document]
jlconlin: Debatable :-D

PR #16, opened 2018-07-20
Feature/sorted

jlconlin:
This branch is the beginning of a namespace used to verify various pieces of the data. In particular, we've created some classes to verify that a range of data is sorted.


apmccartney:
This build breaks coverage. Please revert your changes to the travis files. See the current master for working files

src/ACEtk/details/verify/Sorted/src/check.hpp
```diff
@@ -0,0 +1,15 @@
+template< typename Arg,
+          utility::Require< true, utility::is_range, Range> = true >
+decltype( auto ) check( Arg&& arg ){
+  auto sortedEnd = ranges::is_sorted_until( arg );
+  if( sortedEnd != arg.end() ){
+    auto index = sortedEnd - arg.begin();
```
apmccartney: prefer `std::distance` or `ranges::distance`

src/ACEtk/details/verify/Sorted/src/ctor.hpp
```diff
@@ -0,0 +1,24 @@
...
+        >
+Sorted( Arg&& arg ) :
+  Range( check( std::forward< Arg >( arg ) ) )
+{ }
+
+template< typename... Args, 
+          utility::Require< true, std::is_constructible, Range, Args... > = true
+        >
+Sorted( Args&&... args ):
+  Range( check( std::forward< Args >( args )... ) )
```
apmccartney: This doesn't make sense. Check cannot be applied to multiple arguments. You can only verify sorted-ness after constructing the parent. See the lazy verification in the example I sent.

src/ACEtk/details/verify/test/verify.test.cpp
```diff
@@ -0,0 +1,9 @@
+#define CATCH_CONFIG_MAIN
+#include <chrono>
```
apmccartney: The only header you need here is catch

src/ACEtk/details/verify/src/sorted.hpp
```diff
@@ -0,0 +1,15 @@
+template< typename Range,
+          utility::Require< false, IsSorted, Range > = true,
+          utility::Require< true, utility::is_range, Range > = true 
+        >
+auto sorted( Range&& r ){
+  return Sorted< std::decay_t< Range > >{ std::forward< Range >( r ) };
+}
+
+template< typename Range,
+          utility::Require< false, IsSorted, Range > = true
```
apmccartney: the is only safe if the range is sorted. Here you're enforcing that it's not sorted.

apmccartney (changes requested):
See inline comments. The lack of coverage information is a show stopper.

src/ACEtk/details/verify/test/verify.test.cpp
jlconlin: That must have been copied/pasted from somewhere.

src/ACEtk/details/verify/src/sorted.hpp
jlconlin: Good catch. Another copy/paste error.

jlconlin:
The travis configuration seems to work well. See [here](https://travis-ci.org/njoy/ACEtk/builds/406429555). I did copy from master, with a few minor modifications.

apmccartney:
The travis configuration changes dropped the coverage value, which was working on recent previous pull requests

src/ACEtk/details/verify/Sorted/src/ctor.hpp
apmccartney: Aside from the coverage, I think this is the last thing to be addressed.

jlconlin:
This PR is ready to be looked at (one last time, I hope)

apmccartney (approved):
Excellent work Jeremy

PR #18, opened 2018-07-23
Feature/verify size

jlconlin:


jlconlin:
This Pull Request is ready to go. Waiting for feature/Sorted and feature/Positive will simplify what has to be looked at.

Alternatively, if this one is approved, the other two are included!

src/ACEtk/details/verify/src/size.hpp
```diff
@@ -0,0 +1,10 @@
+template< typename R1, typename R2 >
+inline void equalSize( R1&& a1, R2&& a2 ){
```
apmccartney: `inline` is implicit on function templates. Please remove the keyword.

Here, we ought not mutate `R1` and `R2`. Please accept arguments by const reference.

src/ACEtk/details/verify/src/size.hpp
```diff
@@ -0,0 +1,10 @@
...
+  if( ranges::distance( a1 ) != ranges::distance( a2 ) ){
+    njoy::Log::error( "array sizes are not the same" );
```
apmccartney: This error message is vague. Let's take two `const char*` arguments in the function to specifying names to provide a better error message.

apmccartney (changes requested):

src/ACEtk/details/verify/src/size.hpp
jlconlin: In order to keep this method simpler, I would rather defer more descriptive messages later down the line.

jlconlin: The current way is more consistent with the other verifications in ACEtk. 

I did change the error message a bit.

apmccartney: You'd prefer that, at each call site, you add logic to specifically catch an range-error exception to specify the names of the mismatched ranges? That's an API that's easy to use poorly.

PR #26, opened 2019-04-11
Added OWNER_EXECUTE permissions to install directories.

azukaitis:
My first bug fix for install execute permissions of library when install is requested.


jlconlin:
@azukaitis Thanks for finding this. However, all of our `CMakeLists.txt` files are generated automatically through [metaconfigure](https://github.com/njoy/metaconfigure). This would have to be fixed there.

azukaitis:
thanks!  Just learning about how this works....


Tony

PR #28, opened 2020-10-20
Create ContinuousIntegration.yml

jlconlin:


whaeck (changes requested):
Good for me since this the standard CI file, but you may want to check why the CI fails. Did you run the tests prior to pushing?

jlconlin:
It failed because we don't use variant-adapter anymore.

PR #29, opened 2020-10-20
Updating to C++17

jlconlin:


whaeck (approved):
Since we're going to change the build system anyway, I did not really verify that. Just checked the C++17 flag in the description.json file.

Once CI finishes, you're good to go.

PR #30, opened 2021-07-17
Develop

whaeck:
Updated the CI and the build system in anticipation of ACEtk work.
Removed metaconfigure and travis related files.
Updated .gitignore


cmake/develop_dependencies.cmake
```diff
@@ -0,0 +1,76 @@
...
+    GIT_SHALLOW     TRUE
+    )
+
+FetchContent_Declare( header-utilities
+    GIT_REPOSITORY  https://github.com/njoy/header-utilities
+    GIT_TAG         origin/master
+    GIT_SHALLOW     TRUE
+    )
+
+FetchContent_Declare( range-v3-adapter
```
nathangibson14, line 56: Switch to the real ranges?

nathangibson14 (changes requested):

cmake/develop_dependencies.cmake
whaeck: I was planning on doing that later.

nathangibson14 (approved):

PR #31, opened 2021-07-19
Feature/python

whaeck:
Starting python bindings for ACEtk - currently limited to the generic ACE table.

Updated ACEtk code with documentation, adding header guards, updated tests.

This is the first of a few upcoming pull requests.


david8dixon:
I thought you said this was a big PR - only 72 files changed...

cmake/release_dependencies.cmake
```diff
@@ -45,6 +45,11 @@ FetchContent_Declare( Log
     GIT_TAG         52962b7796afe37ef1d8f7edb4bf9ecb1b868d15
     )
 
+FetchContent_Declare( pybind11
+    GIT_REPOSITORY  https://github.com/pybind/pybind11
+    GIT_TAG         f1abf5d9159b805674197f6bc443592e631c9130 # tag: v2.6.1
+    )
+
 FetchContent_Declare( range-v3-adapter
```
david8dixon, line 53: Is there still a reason for the adapter?

whaeck: That's going to be another pull request (see #32)

david8dixon:
Is there code coverage?

david8dixon:
It doesn't appear to me that the bindings are being built...

whaeck:
No, we switched that off some time ago. We used to get them through travis but our compile times became too high so we abandoned that. I did the coverage reports by hand for a while using gcov but since I was the only one actually doing that, I stopped doing that too. I should look into getting that back up and running though.

whaeck:
There's a separate target named ACEtk.python that is excluded from the make all target.

david8dixon:
ah ok I don't recall that being the case from previous experience with pybind11 but it has been a few years.

david8dixon:
We should really consider removing the ACEtk.python target as a part of this PR. I feel that it would be much better to have it configurable. If I do the following:

```
make ACEtk.python
ctest -N
```

I get a bunch of noise because all the other test executables weren't built.

nathangibson14:
How would you want the Python target to be handled?

Right now, we have it as a separate target, so that ACEtk could be included into a C++ application without the Python hoopla, and yes, the ctest is only set up for the C++ tests.  Our Python test suite is all in unittest, which can be run from a script or from a Python invocation.

Long term, the plan is to move away from building the Python via CMake (at least directly) and move to something like setuptools or pip, as that'd make installation of just the Python quite a bit smoother.

But all that said, we're only pretending we know what we're doing over here, so any advice on the build system is appreciated!

david8dixon:
For now I would make the bindings a cmake option. Maybe have them off by default. Then the python build products and tests can be added conditionally depending on how one configures. This is a somewhat minor detail based on the stumbling around I did initially while trying to build and test.

david8dixon:
Oh and I pushed up a change that has the python unit tests executed with the ctest driver.

whaeck:
As far as I know, the C++ unit tests are already a CMake option that is switched on by default. If we add the python bindings as a CMake option that is also on by default, we can use that option to switch off the bindings in another library that uses ACEtk as a dependency - right?

We still need to do an overhaul of the build system scripts in devtools (one of the things that needs to be done is setting up the CMake python boilerplate automatically) so if you can come up with a proposed solution, that would be great.

Let's see if we can meet next week when we're all onsite to discuss/review a solution.

BTW: this would simplify the CI as well :-)

david8dixon:
Why don’t I create  a branch with a prototype and we can go from there.

whaeck:
@david8dixon Works for me ;-)

david8dixon:
any idea why the checks aren't working?

whaeck:
I've poked the CI to force a build and it seems to work. Did you see something in particular that went wrong?

david8dixon (approved):

david8dixon:
I just wanted to be sure tests passed with the build system changes merged into this branch.

david8dixon:
once this is in I will rebase the ranges branch and review

PR #32, opened 2021-07-21
Update/ranges

whaeck:
Updating the ranges dependencies.


whaeck:
@david-a-dixon You'll get deprecation warnings for the ranges library with this branch. I fixed those in the feature/ce branch so you can ignore those warnings for now.

cmake/develop_dependencies.cmake
```diff
@@ -37,7 +37,7 @@ FetchContent_Declare( disco
 
 FetchContent_Declare( interpolation
     GIT_REPOSITORY  https://github.com/njoy/interpolation
-    GIT_TAG         origin/master
```
david8dixon, line 40: Is having all of these various {build type}_dependencies.cmake files working for you? Hard to tell if they are different.

src/ACEtk/Table/Data/src/XSS.hpp
```diff
@@ -15,7 +15,7 @@ auto XSS( std::size_t index, std::size_t length ){
   #endif
   const auto left = std::next( this->xss_.begin(), index - 1 );
   const auto right = left + length;
-  return ranges::make_iterator_range( left, right );
+  return ranges::make_subrange( left, right );
```
david8dixon, line 18: neat

src/ACEtk/details/verify/Positive/test/Positive.test.cpp
```diff
@@ -46,7 +46,7 @@ SCENARIO( "Testing positivity verification" ){
             pdVector = Positive< std::vector< double > >( posi ) );
 
             pdVector = { 1.0, 2.0, 6.0, 4.0 };
-        CHECK( ranges::equal( { 1.0, 2.0, 6.0, 4.0 }, pdVector ) );
+        //CHECK( ranges::equal( { 1.0, 2.0, 6.0, 4.0 }, pdVector ) );
```
david8dixon, line 49: Why are there tests commented out

src/ACEtk/interpretation/DEDX1/CRTP.hpp
```diff
@@ -10,8 +10,8 @@ class CRTP{
   using TempT = decltype( 1.0 * mev );
 
   #include "ACEtk/interpretation/DEDX1/CRTP/src/outOfBounds.hpp"
-  #include "ACEtk/interpretation/DEDX1/CRTP/S1.hpp"
-  #include "ACEtk/interpretation/DEDX1/CRTP/S0.hpp"
+  //#include "ACEtk/interpretation/DEDX1/CRTP/S1.hpp"
```
david8dixon: If there is a preference to remove the CPT code that doesn't work we could create a branch to explicitly capture it and then clear out whatever we aren't using. That is actually my preference.

david8dixon (approved):

PR #35, opened 2021-08-10
making ACEtk python bindings a configurable option

david8dixon:
Added new cmake option `ACEtk.python` and set to `ON` by default. For users building the code they can:

```
cmake ${src_dir}
make -jN
```

and they will build the python bindings by deafult so `make ACEtk.python` is no longer required. Further, `ACEtk_unit_tests=OFF` by default, so users that are simply after python bindings will no longer build any testing which cuts build times down significantly. For devs, they must configure as follows:

```
cmake -D ACEtk_unit_tests=ON ${src_dir}
make -jN
```
This will build bindings, and build and  configure all tests (including python tests) such that:

```
ctest -jN
```

works as expected. One may also:

```
cmake -D ACEtk.python=OFF -D ACEtk_unit_tests=ON ${src_dir}
make -jN
ctest -jN # this only runs c++ tests because the python tests are not configured when bindings are off
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,15 +1,7 @@
 ########################################################################
 # Preamble
 ########################################################################
-
 cmake_minimum_required( VERSION 3.14 )
-
-if(DEFINED PROJECT_NAME)
-  set(subproject ON)
-else()
-  set(subproject OFF)
-endif()
-
 project( ACEtk LANGUAGES CXX )
 
 ########################################################################
@@ -19,7 +11,7 @@ project( ACEtk LANGUAGES CXX )
 set( CMAKE_CXX_STANDARD 17 )
 set( CMAKE_CXX_STANDARD_REQUIRED YES )
 
-option( ACEtk_unit_tests "Compile the ACEtk unit tests and integrate with ctest" OFF )
+option(ACEtk_unit_tests "Build the ACEtk unit tests" OFF)
 option( ACEtk.python "Build ACEtk python bindings" ON )
 option( strict_compile
     "Treat all warnings as errors." ON
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,14 @@
 ########################################################################
 
 cmake_minimum_required( VERSION 3.14 )
-project( ACEtk LANGUAGES CXX )
 
+if(DEFINED PROJECT_NAME)
+  set(subproject ON)
+else()
+  set(subproject OFF)
+endif()
+
+project( ACEtk LANGUAGES CXX )
 
 ########################################################################
 # Project-wide setup
@@ -13,16 +19,15 @@ project( ACEtk LANGUAGES CXX )
 set( CMAKE_CXX_STANDARD 17 )
 set( CMAKE_CXX_STANDARD_REQUIRED YES )
 
-option( ACEtk_unit_tests
-    "Compile the ACEtk unit tests and integrate with ctest" ON
-    )
+option( ACEtk_unit_tests "Compile the ACEtk unit tests and integrate with ctest" OFF )
+option( ACEtk.python "Build ACEtk python bindings" ON )
 option( strict_compile
     "Treat all warnings as errors." ON
     )
 
 # Compile flags
-set( common_flags "-Wall" "-Wextra" "-Wpedantic" )
-set( strict_flags "-Werror" )
+set( common_flags "-Wall" "-Wextra" "-Wpedantic" ) # i'm not sure why we manage flags this way...
+set( strict_flags "-Werror" )                      # improving this will need to happen in the future for spack
 set( release_flags "-O3" )
 set( debug_flags "-O0" "-g" )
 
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -79,29 +76,26 @@ target_link_libraries( ACEtk
...
     python/src/Table.python.cpp
-)
-target_link_libraries( ACEtk.python PRIVATE ACEtk )
-target_compile_options( ACEtk.python PRIVATE "-fvisibility=hidden" )
-set_target_properties( ACEtk.python PROPERTIES OUTPUT_NAME ACEtk )
-set_target_properties( ACEtk.python PROPERTIES COMPILE_DEFINITIONS "PYBIND11" )
-set_target_properties( ACEtk.python PROPERTIES POSITION_INDEPENDENT_CODE ON)
+  )
+  target_link_libraries( ACEtk.python PRIVATE ACEtk )
+  target_include_directories( ACEtk.python PRIVATE python/src )
```
whaeck, line 86: I like this change, since the includes are only internal to the bindings anyway.

CMakeLists.txt
```diff
@@ -1,28 +1,25 @@
...
+option( ACEtk.python "Build ACEtk python bindings" ON )
 option( strict_compile
     "Treat all warnings as errors." ON
     )
 
 # Compile flags
-set( common_flags "-Wall" "-Wextra" "-Wpedantic" )
-set( strict_flags "-Werror" )
+set( common_flags "-Wall" "-Wextra" "-Wpedantic" ) # i'm not sure why we manage flags this way...
+set( strict_flags "-Werror" )                      # improving this will need to happen in the future for spack
```
nathangibson14, line 22: I took this from our previous build system based on Austin's metaconfigure.  That had a much more robust flag management, with flags set by compiler, etc.  I slimmed it down for now, but we may need more of the old functionality back at some point.

Always happy to see suggestions for improvements :)

Note that these options like strict_compile aren't even hooked up to the build step -- that's a sloppy work-in-progress issue that I haven't gone back to fix.  So for sure... work to be done!

whaeck, line 22: Nathan is going to update the build system at some point to better handle the compilation flags (and other things).

I think this still comes from the old metaconfigure system.

whaeck:
I've tested it and it seems to work, so I'm good. If Nathan is OK with it, we're good to go.

CMakeLists.txt
nathangibson14: The Python includes were a quick kluge Wim put together, I haven't added this to devtools yet.  I'm glad to see you went in the same direction as I was planning.

CMakeLists.txt
```diff
@@ -79,29 +76,26 @@ target_link_libraries( ACEtk
...
+  set_target_properties( ACEtk.python PROPERTIES OUTPUT_NAME ACEtk )
+  set_target_properties( ACEtk.python PROPERTIES COMPILE_DEFINITIONS "PYBIND11" )
+  set_target_properties( ACEtk.python PROPERTIES POSITION_INDEPENDENT_CODE ON)
+endif()
 
 #######################################################################
 # Top-level Only
 #######################################################################
 
 if( CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR )
```
nathangibson14, line 97: Looking at this again, it doesn't make sense to put this in a top-level only block if it's controlled by a CMake variable.  But that's not related to your changes, so I'll just add this as a todo for myself!

cmake/unit_testing.cmake
```diff
@@ -29,4 +29,6 @@ add_subdirectory( src/ACEtk/interpretation/MP1/test )
 #######################################################################
 # python tests 
 #######################################################################
-add_subdirectory( python/test )
+if( ACEtk.python )
+  add_subdirectory( python/test )
```
nathangibson14, line 33: I was a bit confused as to how this works, but it looks like you added a CMakeLists.txt into the Python folder separate from this PR.  I like having this available from CTest for sure.

nathangibson14 (approved):
Looks good to me.  I promise to get back to updating devtools soon and I'll use this sort of pattern as the default for all our repos going forward.  I'll end up changing some of the whitespace from what David used to match my CMake preferences, but I don't think that'll offend anyone  :smile:

whaeck (approved):

CMakeLists.txt
david8dixon: My suggestion would be to use the defaults provided by cmake to begin. We can add custom build types in a `CustomBuildTypes.cmake` file where we capture the extra stuff we have here. The reasoning is that we don't want to break from convention if possible. Especially, once we move towards providing spack recipes because spack will take care of adding platform specific flags like `march and mtune`. 

I agree - this can come later.

david8dixon: BTW not offended at all by formatting preferences - what are they?

david8dixon: Last thing - what is devtools?

david8dixon:
Looks like everyone is good with these changes so I am going to merge and we can continue addressing build system in other PRs as we see fit.

PR #39, opened 2021-09-28
Feature/header

whaeck:
Dealing with the new style headers: skip them and read the normal header instead


nathangibson14 (approved):
Works for me, strategy makes sense.

PR #40, opened 2021-10-04
Adding xsdir components

whaeck:
C++ code for an xsdir entry

Full xsdir to follow, along with python bindings but wanted to make these as small as possible for easy reviewing ;-)

### DIFF
--- a/src/ACEtk.hpp
+++ b/src/ACEtk.hpp
@@ -1,3 +1,6 @@
+// xsdir
+#include "ACEtk/Xsdir.hpp"
+
 // generic ACE table
 #include "ACEtk/State.hpp"
 #include "ACEtk/Table.hpp"
--- a/src/ACEtk/Xsdir/src/ctor.hpp
+++ b/src/ACEtk/Xsdir/src/ctor.hpp
@@ -1,4 +1,4 @@
-Xsdir() = delete;
+Xsdir() = default;
 Xsdir( const Xsdir& ) = default;
 Xsdir( Xsdir&& ) = default;
 
--- a/src/ACEtk/Xsdir/src/ctor.hpp
+++ b/src/ACEtk/Xsdir/src/ctor.hpp
@@ -5,12 +5,13 @@ Xsdir( Xsdir&& ) = default;
 /**
  *  @brief Constructor
  *
- *  @param[in] weights    the atomic weight ratios
+ *  @param[in] ratios     the atomic weight ratios
  *  @param[in] entries    the file name
  */
-Xsdir( std::vector< XsdirEntry > entries,
+Xsdir( std::map< unsigned int, double > ratios,
+       std::vector< XsdirEntry > entries,
        std::optional< std::string > datapath = std::nullopt ) :
-  datapath_( std::move( datapath ) ), awr_(),
+  datapath_( std::move( datapath ) ), awr_( std::move( ratios ) ),
   entries_( std::move( entries ) ), indices_( indices( entries_ ) ) {}
 
 /**
--- a/src/ACEtk/Xsdir/src/ctor.hpp
+++ b/src/ACEtk/Xsdir/src/ctor.hpp
@@ -10,16 +10,15 @@ Xsdir( Xsdir&& ) = default;
  */
 Xsdir( std::vector< XsdirEntry > entries,
        std::optional< std::string > datapath = std::nullopt ) :
-  datapath_( std::move( datapath ) ), entries_( std::move( entries ) ),
-  indices_( indices( entries_ ) ) {}
+  datapath_( std::move( datapath ) ), awr_(),
+  entries_( std::move( entries ) ), indices_( indices( entries_ ) ) {}
 
 /**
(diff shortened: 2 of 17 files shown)


src/ACEtk.hpp
```diff
@@ -1,3 +1,6 @@
+// xsdir
+#include "ACEtk/Xsdir.hpp"
```
nathangibson14, line 2: I don't really like this capitalization, but it also really doesn't matter.

src/ACEtk/Xsdir/src/ctor.hpp
```diff
@@ -0,0 +1,31 @@
+Xsdir() = default;
+Xsdir( const Xsdir& ) = default;
+Xsdir( Xsdir&& ) = default;
```
nathangibson14, line 3: You've previously put this kind of thing in ifdef blocks for Python.  I don't think that's necessary to go back to, but wanted to highlight it.

src/ACEtk/Xsdir/src/parse.hpp
```diff
@@ -0,0 +1,125 @@
+static void trim( std::string string ) {
```
nathangibson14: I'm guessing you would have noticed in testing if this didn't work, but does this really work?  Looks like you are trimming in place but not passing by reference.

src/ACEtk/Xsdir/src/stream.hpp
```diff
@@ -0,0 +1,36 @@
...
+inline std::istream &operator>>( std::istream& in, Xsdir& xsdir ) {
+
+  try {
+
+    Xsdir temp( in );
+    xsdir = std::move( temp );
+  }
+  catch ( ... ) {
+
+    // nothing to do to recover from the error: in is already in fail state
```
nathangibson14: Does an empty catch block act differently than just not having a try/catch at all?

src/ACEtk/Xsdir/test/Xsdir.test.cpp
```diff
@@ -0,0 +1,141 @@
...
+        CHECK( oss.str() == string );
+      } // THEN
+    } // WHEN
+  } // GIVEN
+} // SCENARIO
+
+std::string chunk() {
+
+  return
+    "datapath=/some/path/to/Data\n"
```
nathangibson14, line 183: Test sometime without a datapath specified?  Test something that needs toLower?

src/ACEtk/XsdirEntry.hpp
```diff
@@ -0,0 +1,135 @@
...
+   */
+  const std::string& fileName() const { return this->filename_; }
+
+  /**
+   *  @brief Return the access route
+   */
+  const std::optional< std::string >& accessRoute() const { return this->access_; }
+
+  /**
+   *  @brief Return the access route
```
nathangibson14: C&P error

nathangibson14 (changes requested):
Looks good overall.  Only real comment is I want to see testing showing that the trim() function and files without datapaths work as expected.

src/ACEtk.hpp
whaeck: I don't like it either but this is how we name our classes everywhere and I wanted to stay consistent. An alternative would we to name it DataDirectoryFile but everybody knows what an xsdir is, which is why I choose that name. What we can do, is create an XSDIR alias, like we do in ENDFtk with the TAB1, CONT, etc.

src/ACEtk/Xsdir/src/ctor.hpp
whaeck: That's only the default constructor. In this case, I needed the default constructor to get it to work properly with the streaming operator (I should check that though). With ENDFtk, I deliberately made the default constructor off limits in the C++ interface because we do not want people to create empty and invalid objects. We may want to go away from that since we know most people will use the python interface, in which the default constructor is not exposed anyway.

src/ACEtk/Xsdir/src/stream.hpp
whaeck: Streaming is particular in that it does not throw exceptions, it actually changes the stream state which a user can then check (this is how stream operators in the standard library work). I catch any exception here to mimic that behaviour. The stream state is already set to fail in the constructor that threw the exception, but I might just do it here as well to make it clear that we do this to mimic the standard library behaviour.

src/ACEtk/Xsdir/src/parse.hpp
whaeck: I'm an idiot. Adding tests with upper case keywords and without a datapath exposed this issue as well.

src/ACEtk/Xsdir/test/Xsdir.test.cpp
whaeck: Done.

src/ACEtk/XsdirEntry.hpp
whaeck: done.

nathangibson14 (approved):

PR #41, opened 2021-10-11
Python bindings for xsdir components

whaeck:
Python bindings for the xsdir components.

Pull request #40 must be merged first


python/src/Xsdir.python.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+  .def_static(
+
+    "from_file",
+    [] ( const std::string& filename ) -> Component {
+
+      std::ifstream in( filename );
+      if ( not in ) {
+
+        njoy::Log::error( "Could not open file \'{}\'", filename );
```
nathangibson14: Why not put this in the C++?

nathangibson14 (commented):
This one looks fine, but I'll wait for the previous one to get merged to approve.

python/src/Xsdir.python.cpp
whaeck: got a point there. Just a sec ...

nathangibson14 (approved):

PR #42, opened 2021-10-14
AND block components

whaeck:
This implements the following AND block components:
- tabulated angular distirbution for a single incident energy
- 32 equiprobable bins for a single incident energy
- isotropic distirbution for a single incident energy
- angular distribution data for a single reaction

This also adds an enum for the angular distribution type.

The AND block is next.


src/ACEtk/block/AngularDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,33 @@
...
+AngularDistributionData( std::vector< Distribution >&& distributions,
+                         std::size_t locb = 1 ) :
+  Base( "AND::AngularDistributionData",
+        generateXSS( std::move( distributions ), locb ) ),
+  locb_( locb ) {}
+
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] lsig    the begin iterator of the LSIG block in the XSS array
```
nathangibson14: Documentation doesn't apply to this constructor.

src/utility/overload.hpp
```diff
@@ -0,0 +1,32 @@
+#ifndef NJOY_UTILITY_OVERLOAD
+#define NJOY_UTILITY_OVERLOAD
```
nathangibson14, line 2: This looks like something taken from another component.  Is this C&P from elsewhere or new?  Does it belong in ACEtk rather than a dependency?

nathangibson14 (changes requested):
One question and one bad piece of documentation.  Everything else looks good.

src/ACEtk/block/AngularDistributionData/src/ctor.hpp
whaeck: Gotta hate copy/paste ...
Should be fixed now.

src/utility/overload.hpp
whaeck: Eventually it should go into a dependency, yes (which is why I already put it separately from ACEtk). When we overhaul header-utilities we might add it to that. It's essentially a very simple piece of template code.

nathangibson14 (approved):

PR #43, opened 2021-10-17
Binding the AND block in darkness

whaeck:
Finished the AND block, the components PR (see #42) needs to be merged first


src/ACEtk/block/AngularDistributionData.hpp
```diff
@@ -169,19 +169,25 @@ class AngularDistributionData : protected Base {
...
+      auto right = this->end();
+      for ( auto next = index + 1; next <= this->numberIncidentEnergies(); ++next ) {
+
+        auto nextlocator = this->relativeAngularDistributionLocator( next );
+        if ( nextlocator > locator ) {
+
+          right = std::next( this->begin(), nextlocator - 1 );
+          break;
+        }
+      }
```
nathangibson14, line 190: What's the story on this one?  Are locators not monotonically increasing?

src/ACEtk/block/DistributionGivenElsewhere.hpp
```diff
@@ -0,0 +1,36 @@
+#ifndef NJOY_ACETK_BLOCK_DISTRIBUTIONGIVENELSEWHERE
```
nathangibson14, line 1: This is an impressive number of lines of code for doing literally nothing.  Gotta love "given elsewhere"!

nathangibson14 (approved):

src/ACEtk/block/AngularDistributionData.hpp
whaeck: This is another example of why you should not store multiple pieces of information in a single field. When the locator is zero it means the distribution is isotropic so it is possible to have the following locators:
```
1   24    0     0   48
``` 
The sub-block that starts at 24 for the second incident energy stops at position 48 - 1, which is just before the data for the 4th incident energy value. I therefore need to go over the different locators and find the first one that is either not zero or larger than the current locator to know where my sub-block stops.

Technically speaking this should NEVER happen since isotropic distributions only happen for low incident energies after which they become anisotropic. However, if it would happen, it would cause an error in ACEtk hence the code protecting itself against a very unlikely situation.

src/ACEtk/block/DistributionGivenElsewhere.hpp
whaeck: I needed a way to represent a reaction with data given in the DLW block :-)

PR #44, opened 2021-10-20
Feature/dlw components

whaeck:
Some refactoring was done on the coding already there so I've put up this one to get that done. The only DLW component in here is the LevelScatteringDistribution.

As usual: the AND components and AND pull request need to be done first


src/ACEtk/block/LevelScatteringDistribution.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+public:
+
+  /* constructor */
+  #include "ACEtk/block/LevelScatteringDistribution/src/ctor.hpp"
+
+  /**
+   *  @brief Return the minimum incident energy for the distribution
+   */
+  double minimumIncidentEnergy() const { return this->emin_; }
```
nathangibson14, line 43: Are `emin` and `emax` specified ACE speak, justifying additional interface?  Or do these not have an official ACE name?

src/ACEtk/block/LevelScatteringDistribution/src/verifySize.hpp
```diff
@@ -0,0 +1,12 @@
+static void verifySize( Iterator begin, Iterator end ) {
+
+  // there are at least 2 values in the block
```
nathangibson14, line 3: Is "at least" the right check?  Isn't this an "exactly" case?

nathangibson14 (commented):
This is probably good to go, I just left a few questions.  I'll approve once answered.

src/ACEtk/block/LevelScatteringDistribution.hpp
whaeck: The minimum and maximum energy are not part of the data for the law (there's only C1 and C2 that are needed). However, these energy values need to be given so that we can generate the obligatory probability = 1 when the law is the only one given. It's part of the abstraction that allows us to omit that probability when serving up data.

As such, they do not have an official ACE name, but we could give them one to make it consistent and that might be the better choice. All other DLW components (see feature/dlw-components-too) have the same interface functions as well.

src/ACEtk/block/LevelScatteringDistribution/src/verifySize.hpp
whaeck: The XSS array can have "gaps" in, hence the "at least". The way the ACE file is set up also allows for gaps to appear. Technically speaking the blocks may even appear anywhere in the file since the starting index is only given as a locator (without an ending one). When generating the blocks, I probably should sort the locators first instead of assuming that they always appear in a specific order. 

Also, NJOY2016 sometimes tweaks the ACE file (I put some of these in actually) and in some cases, we would have to rearrange the XSS array and every single locator in it so I left it as a gap. MCNP doesn't even notice. NJOY2016 does signal whenever a gap is detected during locator checking.

nathangibson14 (approved):

PR #46, opened 2021-10-26
Feature/dlw components too

whaeck:
Added components from the DLW block:
- outgoing energy distributions (ACE LAW=4)
- Kalbach-Mann systematics (ACE LAW=44)

As usual, some stuff comes first ;-)


python/src/block/KalbachMannDistributionData.python.cpp
```diff
@@ -0,0 +1,131 @@
+// system includes
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+// local includes
+#include "ACEtk/block/KalbachMannDistributionData.hpp"
+#include "views.hpp"
```
cjosey, line 7: [clang-tidy - low priority]
Suggest sorting includes alphabetically.

(I note this is consistent with most other implementations - will only mention here)

python/src/block/KalbachMannDistributionData.python.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include "definitions.hpp"
+
+// namespace aliases
+namespace python = pybind11;
+
+namespace block {
+
+void wrapKalbachMannDistributionData( python::module& module,
+                                         python::module& ) {
```
cjosey, line 16: [clang-tidy - recommended]
Add `/* unused */` or equivalent for unused arguments.

(I note this is consistent with most other implementations - will only mention here)

python/src/block/TabulatedKalbachMannDistribution.python.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+    python::init< double, int, std::vector< double >&&, std::vector< double >&&,
+                  std::vector< double >&&, std::vector< double >&&,
+                  std::vector< double >&&, std::size_t >(),
+    python::arg( "incident" ), python::arg( "interpolation" ),
+    python::arg( "cosines" ), python::arg( "pdf" ), python::arg( "cdf" ),
+    python::arg( "r" ), python::arg( "a" ), python::arg( "discrete" ) = 0,
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self        the block\n"
```
cjosey, line 43: [question]
Does the `self` argument need calling out since you're only ever going to really call it as `dist.init()`?

(question applies elsewhere too, but will only be mentioned here)

python/test/CMakeLists.txt
```diff
@@ -15,3 +15,7 @@ add_test( NAME ACEtk.python.block.IsotropicAngularDistribution COMMAND ${PYTHON_
 add_test( NAME ACEtk.python.block.AngularDistributionData COMMAND ${PYTHON_EXECUTABLE} -m unittest -v test/block/Test_ACEtk_AngularDistributionData.py WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/python )
 add_test( NAME ACEtk.python.block.AngularDistributionBlock COMMAND ${PYTHON_EXECUTABLE} -m unittest -v test/block/Test_ACEtk_AngularDistributionBlock.py WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/python )
 add_test( NAME ACEtk.python.block.LevelScatteringDistribution COMMAND ${PYTHON_EXECUTABLE} -m unittest -v test/block/Test_ACEtk_LevelScatteringDistribution.py WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/python )
+add_test( NAME ACEtk.python.block.TabulatedOutgoingEnergyDistribution COMMAND ${PYTHON_EXECUTABLE} -m unittest -v test/block/Test_ACEtk_TabulatedOutgoingEnergyDistribution.py WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/python )
```
cjosey, line 18: [unrelated to this PR - future work]
I would suggest the following:
```cmake
function(add_python_test name file)
  add_test(
    NAME ACEtk.python.block.${name}
    COMMAND ${CMAKE_COMMAND} -E env PYTHONPATH=${CMAKE_BINARY_DIR} ${PYTHON_EXECUTABLE} -m unittest -v test/block/${file}
    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/python
  )
endfunction()

add_python_test(OutgoingEnergyDistributionData Test_ACEtk_OutgoingEnergyDistributionData.py)
[etc]
```

The CMake command `-E env` command allows the `PYTHONPATH` to be set in a system-independent way, and skips a developer needing to set the value themselves to run tests. At some point I need to figure out how our other codes do this, since those don't need to contort this way.

python/test/block/Test_ACEtk_AngularDistributionBlock.py
```diff
@@ -13,8 +13,8 @@
 from ACEtk import TabulatedAngularDistribution
```
cjosey, line 13: [unrelated to this PR]
`from ACEtk import EquiprobableAngularBins` (above) appears to be unused in this test.

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
```diff
@@ -0,0 +1,131 @@
...
+            self.assertAlmostEqual( 1.219437E+01, chunk.incident_energies[0] )
+            self.assertAlmostEqual( 20., chunk.incident_energies[1] )
+
+            self.assertAlmostEqual( 1.219437E+01, chunk.incident_energy(1) )
+            self.assertAlmostEqual( 20., chunk.incident_energy(2) )
+
+            self.assertEqual( 27, chunk.LOCC(1) );
+            self.assertEqual( 39, chunk.LOCC(2) );
+            self.assertEqual( 27, chunk.distribution_locator(1) );
+            self.assertEqual( 39, chunk.distribution_locator(2) );
```
cjosey, line 45: [verification]
Is these being 20 higher than the relative ones due to setting `locb = 21`?

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
```diff
@@ -0,0 +1,131 @@
...
+            self.assertEqual( 7, chunk.relative_distribution_locator(1) );
+            self.assertEqual( 19, chunk.relative_distribution_locator(2) );
+
+            self.assertEqual( True, isinstance( chunk.distribution(1), TabulatedKalbachMannDistribution )  )
+            self.assertEqual( True, isinstance( chunk.distribution(2), TabulatedKalbachMannDistribution ) )
+
+            data1 = chunk.distribution(1);
+            self.assertAlmostEqual( 1.219437E+01, data1.incident_energy )
+            self.assertEqual( 1, data1.interpolation )
+            self.assertEqual( 0, data1.number_discrete_photon_lines )
```
cjosey, line 56: [verification]
It does not appear there's a test for a non-zero number of photon lines. Is this because the format doesn't support it, but it still provides the interface?

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
```diff
@@ -0,0 +1,131 @@
...
+            self.assertAlmostEqual( 0., data1.outgoing_energies[0] )
+            self.assertAlmostEqual( 1.866919E-02, data1.outgoing_energies[-1] )
+
+            self.assertEqual( 2, len( data1.pdf ) )
+            self.assertAlmostEqual( 5.356419E+01, data1.pdf[0] )
+            self.assertAlmostEqual( 0., data1.pdf[-1] )
+
+            self.assertEqual( 2, len( data1.cdf ) )
+            self.assertAlmostEqual( 0., data1.cdf[0] )
+            self.assertAlmostEqual( 1., data1.cdf[-1] )
```
cjosey, line 69: [Question]
I know that these are floating point values and it is best-practice for these to be checked this way, but both 0 and 1 are exactly represent-able in IEEE-floats. Are descendant applications expected to handle them not quite being 0/1?

Similarly, are there any chances of inconsistent precision, where the float64 values in the python side do not precisely match the internal representation of ACEtk such that the non-zero/one values elsewhere would not assertEqual?

python/test/block/Test_ACEtk_TabulatedKalbachMannDistribution.py
```diff
@@ -0,0 +1,73 @@
...
+    def test_component( self ) :
+
+        def verify_chunk( self, chunk ) :
+
+            # verify content
+            self.assertEqual( False, chunk.empty )
+            self.assertEqual( 17, chunk.length )
+            self.assertEqual( "DLW::TabulatedKalbachMannDistribution", chunk.name )
+
+            self.assertEqual( 2.1, chunk.incident_energy )
```
cjosey, line 27: [Question]
The chunk above does not contain 2.1. Is this just something not part of the chunk and created below?

python/test/block/Test_ACEtk_TabulatedKalbachMannDistribution.py
```diff
@@ -0,0 +1,73 @@
...
+            self.assertEqual( 3, chunk.number_discrete_photon_lines )
+            self.assertEqual( 3, chunk.number_outgoing_energies )
+
+            self.assertEqual( 3, len( chunk.outgoing_energies ) )
+            self.assertAlmostEqual( 1e-11, chunk.outgoing_energies[0] )
+            self.assertAlmostEqual( 20., chunk.outgoing_energies[-1] )
+
+            self.assertEqual( 3, len( chunk.pdf ) )
+            self.assertAlmostEqual( .5, chunk.pdf[0] )
+            self.assertAlmostEqual( .5, chunk.pdf[-1] )
```
cjosey, line 38: [Question]
Are there any concerns in behavior that this PDF is not normalized?  If so, should the library check for it?

python/test/block/Test_ACEtk_TabulatedOutgoingEnergyDistribution.py
```diff
@@ -0,0 +1,61 @@
...
+            xss = chunk.xss_array
+            for index in range( chunk.length ) :
+
+                self.assertAlmostEqual( self.chunk[index], xss[index] )
+
+        # the data is given explicitly
+        chunk = TabulatedOutgoingEnergyDistribution(
+                    incident = 2.1,
+                    interpolation = 2,
+                    cosines = [ 1e-11, 1., 20. ],
```
cjosey, line 52: These are energies, not cosines.

python/test/block/Test_ACEtk_TabulatedKalbachMannDistribution.py
```diff
@@ -0,0 +1,73 @@
...
+            xss = chunk.xss_array
+            for index in range( chunk.length ) :
+
+                self.assertAlmostEqual( self.chunk[index], xss[index] )
+
+        # the data is given explicitly
+        chunk = TabulatedKalbachMannDistribution(
+                    incident = 2.1,
+                    interpolation = 2,
+                    cosines = [ 1e-11, 1., 20. ],
```
cjosey, line 62: These are energies, not cosines.

src/ACEtk/block/AngularDistributionData.hpp
```diff
@@ -118,7 +118,7 @@ class AngularDistributionData : protected details::Base {
    *
    *  @param[in] index     the index (one-based)
    */
-  int angularDistributionLocator( std::size_t index ) const {
+  int distributionLocator( std::size_t index ) const {
```
cjosey, line 121: [clang-tidy - low priority]
For constant functions with no side effects, add `[[nodiscard]]`.

(Won't bring up on future examples)

src/ACEtk/block/KalbachMannDistributionData.hpp
```diff
@@ -0,0 +1,142 @@
...
+#include <variant>
+
+// other includes
+#include "utility/overload.hpp"
+#include "ACEtk/block/details/BaseDistributionData.hpp"
+#include "ACEtk/block/TabulatedKalbachMannDistribution.hpp"
+
+namespace njoy {
+namespace ACEtk {
+namespace block {
```
cjosey, line 14: [clang-tidy, low priority]
Can collapse to `namespace njoy::ACEtk::block {`

(Won't mention on further examples)

src/ACEtk/block/KalbachMannDistributionData.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+public:
+
+  /* constructor */
+  #include "ACEtk/block/KalbachMannDistributionData/src/ctor.hpp"
+
+  /**
+   *  @brief Return the number of incident energy values
+   */
+  std::size_t NE() const { return BaseDistributionData::NE(); }
```
cjosey, line 36: [Question]
Can these simple function wrappers all not be replaced with:
```c++
using BaseDistrbutionData::NE;
```
or similar? Is there a specific improvement aside from adding doxygen comments on top?

src/ACEtk/block/KalbachMannDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,46 @@
+KalbachMannDistributionData() = default;
```
cjosey, line 1: [clang-tidy - recommended]
All constructors are defined, but there is no destructor.

src/ACEtk/block/KalbachMannDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,46 @@
...
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] boundaries       the interpolation range boundaries
+ *  @param[in] interpolants     the interpolation types for each range
+ *  @param[in] distributions    the distributions for each incident energy
+ *  @param[in] locb             the starting xss index with respect to the superblock
+ */
+KalbachMannDistributionData(
+    std::vector< long >&& boundaries,
```
cjosey, line 15: [clang-tidy - low priority]
`long`'s behavior is not portable in that it is 32-bits on Windows. Suggest replacing `long` with `int64_t` and `short` with `int16_t` elsewhere (or the `_least` equivalents if you want to be portable for weird 72-bit machines).

src/ACEtk/block/KalbachMannDistributionData/test/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+
+add_executable( ACEtk.block.KalbachMannDistributionData.test KalbachMannDistributionData.test.cpp )
```
cjosey, line 2: [Suggestion, future work]
I would suggest similar to the python stuff to extract out this boilerplate into a test-making function.

src/ACEtk/block/OutgoingEnergyDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,44 @@
+OutgoingEnergyDistributionData() = default;
```
cjosey, line 1: [clang-tidy - recommended]
All constructors are defined, but there is no destructor.

src/ACEtk/block/TabulatedKalbachMannDistribution.hpp
```diff
@@ -0,0 +1,120 @@
...
+    return TabulatedProbabilityDistribution::interpolation() % 10;
+  }
+
+  /**
+   *  @brief Return the number of discrete photon lines
+   */
+  std::size_t numberDiscretePhotonLines() const {
+
+    // the INTT' value = 10 * ND + INTT
+    return ( TabulatedProbabilityDistribution::interpolation() -
```
cjosey, line 56: _Technically_ since integer division floors the result and all arguments are non-negative, the subtraction isn't necessary.

However, any change is irrelevant. The compiler will optimize this all out and there won't even be a division operator left. So whatever is most readable.

src/ACEtk/block/TabulatedKalbachMannDistribution/src/ctor.hpp
```diff
@@ -0,0 +1,45 @@
+TabulatedKalbachMannDistribution() = default;
```
cjosey, line 1: [clang-tidy - recommended]
All constructors are defined, but there is no destructor.

src/ACEtk/block/TabulatedKalbachMannDistribution/src/ctor.hpp
```diff
@@ -0,0 +1,45 @@
...
+
+TabulatedKalbachMannDistribution( const TabulatedKalbachMannDistribution& ) = default;
+TabulatedKalbachMannDistribution( TabulatedKalbachMannDistribution&& ) = default;
```
cjosey, line 4: [clang-tidy]
Apparently, this constructor is capable of throwing.

src/ACEtk/block/TabulatedKalbachMannDistribution/src/ctor.hpp
```diff
@@ -0,0 +1,45 @@
...
+                                  int interpolation,
+                                  std::vector< double >&& cosines,
+                                  std::vector< double >&& pdf,
+                                  std::vector< double >&& cdf,
+                                  std::vector< double >&& r,
+                                  std::vector< double >&& a,
+                                  std::size_t discrete = 0 ) :
+  TabulatedProbabilityDistribution(
+        "DLW::TabulatedKalbachMannDistribution",
+        discrete * 10 + interpolation, std::move( cosines ),
```
cjosey, line 28: [clang-tidy - recommended]
`discrete` is `size_t`, destination is `int`.  Suggest making consistent.

src/ACEtk/block/TabulatedOutgoingEnergyDistribution/src/ctor.hpp
```diff
@@ -0,0 +1,42 @@
+TabulatedOutgoingEnergyDistribution() = default;
```
cjosey, line 1: [clang-tidy - recommended]
All constructors are defined, but there is no destructor.

src/ACEtk/block/details/BaseDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,48 @@
+BaseDistributionData() = default;
```
cjosey, line 1: [clang-tidy - recommended]
All constructors are defined, but there is no destructor.

src/ACEtk/block/TabulatedOutgoingEnergyDistribution.hpp
```diff
@@ -0,0 +1,103 @@
...
+ *
+ *  The TabulatedOutgoingEnergyDistribution class contains the probability
+ *  density function (PDF) and cumulative density function (CDF) as a function
+ *  of the outgoing energy for a single incident energy.
+ */
+class TabulatedOutgoingEnergyDistribution :
+  protected details::TabulatedProbabilityDistribution {
+
+  /* fields */
+  double incident_;
```
cjosey, line 26: Not all constructors guarantee this is set. Suggest adding a default value.

src/ACEtk/block/TabulatedOutgoingEnergyDistribution/src/ctor.hpp
```diff
@@ -0,0 +1,42 @@
...
+ */
+TabulatedOutgoingEnergyDistribution( double incident,
+                                     int interpolation,
+                                     std::vector< double >&& cosines,
+                                     std::vector< double >&& pdf,
+                                     std::vector< double >&& cdf,
+                                     std::size_t discrete = 0 ) :
+  TabulatedProbabilityDistribution(
+        "DLW::TabulatedOutgoingEnergyDistribution",
+        discrete * 10 + interpolation, std::move( cosines ),
```
cjosey, line 24: [clang-tidy - recommended]
discrete is `size_t`, destination is `int`. Suggest making consistent.

src/ACEtk/block/details/BaseDistributionData.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ *  @class
+ *  @brief Base class for distribution data as a function of incident energy
+ */
+template< typename Distribution >
+class BaseDistributionData : protected details::Base {
+
+  /* fields */
+  std::size_t locb_;
```
cjosey, line 26: Not all constructors initialize this value. Suggest putting 0.

src/ACEtk/block/details/BaseDistributionData.hpp
```diff
@@ -0,0 +1,193 @@
...
+
+public:
+
+  /* constructor */
+  #include "ACEtk/block/details/BaseDistributionData/src/ctor.hpp"
+
+  /**
+   *  @brief Return the interpolation data
+   */
+  auto interpolationData() const { return this->interpolation_; }
```
cjosey, line 44: [unsure]
Would it be better to return this as `auto &`?

src/ACEtk/block/details/BaseDistributionData.hpp
```diff
@@ -0,0 +1,193 @@
...
+   *  (debug mode only).
+   *
+   *  @param[in] index     the index (one-based)
+   */
+  int LOCC( std::size_t index ) const {
+
+    #ifndef NDEBUG
+    this->verifyIncidentEnergyIndex( index );
+    #endif
+    return this->data_.value( 2, index );
```
cjosey, line 123: [low priority, style]
Return type of `this->data_.value` is `double`. Suggest `static_cast<int>`?

src/ACEtk/block/details/BaseDistributionData.hpp
```diff
@@ -0,0 +1,193 @@
...
+   *
+   *  When the index is out of range an std::out_of_range exception is thrown
+   *  (debug mode only).
+   *
+   *  @param[in] index     the index (one-based)
+   */
+  std::size_t relativeDistributionLocator( std::size_t index ) const {
+
+    const int locator = this->LOCC( index );
+    return locator - this->locb_ + 1;
```
cjosey, line 157: [design question]
The locator is an integer, yet we're returning a size_t. Is there a reason for the conversion?

src/ACEtk/block/details/BaseDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,48 @@
...
+ *  @param[in] name             the name of the block
+ *  @param[in] locb    the starting xss index with respect to the superblock
+ *  @param[in] sig     the begin iterator of the block in the XSS array
+ *  @param[in] end     the end iterator of the block in the XSS array
+ */
+BaseDistributionData( std::string&& name, std::size_t locb,
+                      Iterator begin, Iterator end ) :
+  Base( std::move( name ), begin, end ), locb_( locb ) {
+
+  std::size_t nr = static_cast< std::size_t >( this->XSS( 1 ) );
```
cjosey, line 41: [clang-tidy, recommended]
`nr` and `ne` can be marked const. Also, since the type is declared in the `static_cast`, suggest just making these `const auto`.

src/ACEtk/block/details/BaseDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,48 @@
...
+  std::size_t ne = static_cast< std::size_t >( this->XSS( 1 + 2 * nr + 1 ) );
+  verifySize( this->begin(), this->end(), nr, ne );
+  this->generateBlocks();
+}
+
+BaseDistributionData& operator=( const BaseDistributionData& ) = default;
+BaseDistributionData& operator=( BaseDistributionData&& ) = default;
```
cjosey, line 48: [clang-tidy - low-priority]
It is suggesting marking all move constructors as noexcept.

src/ACEtk/block/details/BaseDistributionData/src/generateBlocks.hpp
```diff
@@ -0,0 +1,13 @@
+void generateBlocks() {
+
+  std::size_t nr = static_cast< std::size_t >( this->XSS( 1 ) );
```
cjosey, line 3: Same comments on `nr` and `ne` below.

src/ACEtk/block/details/BaseDistributionData/src/generateBlocks.hpp
```diff
@@ -0,0 +1,13 @@
...
+  std::size_t ne = static_cast< std::size_t >( this->XSS( 1 + 2 * nr + 1 ) );
+  auto begin = this->begin();
+  auto data = begin + 2 * nr + 1;
+  auto end = data + + 2 * ne + 1;
```
cjosey, line 7: Remove second + in `data +  + 2`.

src/ACEtk/block/details/BaseDistributionData/src/generateBlocks.hpp
```diff
@@ -0,0 +1,13 @@
+void generateBlocks() {
+
+  std::size_t nr = static_cast< std::size_t >( this->XSS( 1 ) );
+  std::size_t ne = static_cast< std::size_t >( this->XSS( 1 + 2 * nr + 1 ) );
```
cjosey, line 4: [suggestion - complexity]
The entry in the ACE format manual is simply `2 + 2*nr`.

src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+    // insert the xss array
+    xss.insert( xss.end(), distribution.begin(), distribution.end() );
+
+    // on to the next distribution
+    ++index;
+  }
+
+  // insert the interpolation data in the front
+  InterpolationData interpolation( std::string( name ),
```
cjosey, line 31: [clang-tidy - recommended]
Can be marked const. Same with `nr` and `ne` above.

src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
```diff
@@ -0,0 +1,37 @@
...
+             std::vector< long >&& boundaries,
+             std::vector< long >&& interpolants,
+             std::vector< Distribution >&& distributions,
+             std::size_t locb ) {
+
+  // add the distribution data to the xss array
+  std::size_t nr = boundaries.size();
+  std::size_t ne = distributions.size();
+  std::vector< double > xss( 1 + 2 * ne );
+  xss[0] = ne;
```
cjosey, line 12: Suggest `static_cast<double>(ne)`. Same with other int-to-xss operations elsewhere.

src/ACEtk/block/details/TabulatedProbabilityDistribution/src/generateXSS.hpp
```diff
@@ -2,14 +2,17 @@ static std::vector< double > generateXSS(
...
     std::vector< double >&& pdf,
-    std::vector< double >&& cdf ) {
+    std::vector< double >&& cdf,
+    std::vector< std::vector< double > >&& data ) {
 
   std::vector< double > xss = { static_cast< double >( interpolation ) };
+  data.insert( data.begin(), std::move( cdf ) );
+  data.insert( data.begin(), std::move( pdf ) );
+  data.insert( data.begin(), std::move( values ) );
   xss.reserve( 2 + 3 * values.size() );
```
cjosey, line 12: Move this reserve statement above the `values` `std::move` statement.

cjosey (changes requested):
Alright. So, there are a few notes here.

Terms marked `[clang-tidy]` mean that that tool told me about these. If they were low priority, I noted the first time they came up only so that you can tell me if you want to persist in future reviews with these checks. The recommended ones I noted down every time they came up.

There were two concerning places. First, there was using the `cosines` in the Python to set the energy values. The second was using an array after it had been `move`d. Addressing those (or telling me I'm wrong) are what block approval.

The rest is suggestions and entirely up to you.

python/src/block/TabulatedKalbachMannDistribution.python.cpp
whaeck: When I started putting in documentation in ENDFtk, I added self as an argument in the documentation since that's what I had done when I wrote python code. As a result, we continue doing this for consistency reasons.

python/test/CMakeLists.txt
whaeck: Tim has proposed a few changes in the build system and I think something like this is already in there. Revamping the build system will be a priority before we make a mew release in ENDFtk (and we'll do all NJOY21 repos at that point).

python/test/block/Test_ACEtk_AngularDistributionBlock.py
whaeck: Since these PRs are way behind, I'm making the change in the fix/review-1 branch.

Nice catch, I don;t go this far in reviews. Impressive. Very impressive.

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
whaeck: Yes. That is indeed the idea. Since locators are relative to the beginning of the superblock they are in instead of the beginning of the actual block they are in, I need to put in this locb nonsense. A user won't have to deal with it though since the user can just make the block with the default locb value (being 0). Internally, any block that has locb issues gets recreated whenever these internal locators have to change.

This is one of the things I would like to change once we go to a newer ACE format that keeps a flat array: all locators are relative to the beginning of the block they are defined in.

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
whaeck: This is an outstanding question: see issue #69 

I think the specs allow for it but I've never seen it used. I prefer to err on the edge of caution before removing the photon line interface all together.

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
whaeck: There are spots where we have small differences, yes. Whether those are representation issues - or read issues in ACEtk I do not know. Tim did tell me that the disco routines we use to read the ASCII data are always off by a single bit. This is why 1e-11 sometimes comes out as 9.9999999...e-12 instead of 1e-11 exactly. Using strtod() instead of disco seems to fix that. It is an issue I will have to look into at some point.

python/test/block/Test_ACEtk_TabulatedKalbachMannDistribution.py
whaeck: The incident energy is stored elsewhere in the XSS array and does not go into the xss chunk associated to this block. The XSS array for a reaction looks like this:

E_i,1 ... E_i,NE
LOC_i,1 ... LOC_i,NE
data for E_i,1 at LOC_i,1
...
data for E_i,NE at LOC_i,NE

To simplify the interface of the constructor, I require the user to give me the incident energy value for each distribution. Hence the 2.1 not being in the xss chunk.

python/test/block/Test_ACEtk_TabulatedKalbachMannDistribution.py
whaeck: It's a valid question but the answer is I don't really know.

We want ACEtk to read, write and modify ACE files so I might make sense. On the other hand, we don;t want to spend our time testing the same file over and over again every time we open a file. It's an open issue.

python/test/block/Test_ACEtk_TabulatedOutgoingEnergyDistribution.py
whaeck: This is one of those cases where the same block is used because they are exactly the same in structure but slightly different in interface.

This is the documentation that goes with it:
```
 *  The TabulatedEnergyDistribution class contains the probability
 *  density function (PDF) and cumulative density function (CDF) as a function
 *  of the outgoing energy for the given incident energy or cosine value. It is
 *  used in the OutgoingEnergyDistributionData (ACE LAW 4 for a given incident
 *  energy) and in the OutgoingAngleEnergyDistributionData (ACE LAW 67 for a
 *  given cosine) in the DLW block.
```

I'm not sure if a single keyword difference merits an additional class?

python/test/block/Test_ACEtk_TabulatedKalbachMannDistribution.py
whaeck: Same as above.

src/ACEtk/block/AngularDistributionData.hpp
whaeck: You'll have to explain to me what this does.

src/ACEtk/block/KalbachMannDistributionData.hpp
whaeck: I honestly don't remember. I did refactor this later. BaseDistributionData is a more generic block so NE() on it is now N() - requiring me to rename the function anyway.

src/ACEtk/block/KalbachMannDistributionData/src/ctor.hpp
whaeck: Yeah, I'm not very consistent on this. Even ENDFtk is a mess when it comes to these things.

I might clean this up in a single go though.

src/ACEtk/block/TabulatedKalbachMannDistribution.hpp
whaeck: I'll leave it as is.

src/ACEtk/block/TabulatedKalbachMannDistribution/src/ctor.hpp
whaeck: I'll leave this for a separate change.

src/ACEtk/block/TabulatedOutgoingEnergyDistribution.hpp
whaeck: Right. This should be only an issue with the default constructor, right?

I only define those because the python bindings need a default constructor when a variant is used (in ENDFtk I used to wrap these in precompiler statements but I didn't do that here).

src/ACEtk/block/details/BaseDistributionData.hpp
whaeck: I changed this to const InterpolationData&

auto makes people lazy ...

src/ACEtk/block/details/BaseDistributionData.hpp
whaeck: Yes, but casting is not the right thing to do. Rounding is better.

This was changed in a later PR (I think it is in the fix/rounding one). It now says the following:
```
int LOCC( std::size_t index ) const {

    return round( this->value( this->values_.NC(), index ) );
  }
```

src/ACEtk/block/details/BaseDistributionData.hpp
whaeck: Me sloppy with integer types.

src/ACEtk/block/details/BaseDistributionData/src/ctor.hpp
whaeck: I'm not being consistent. I actually do this in other locations. Changed this in the fix/review-1 branch

src/ACEtk/block/details/BaseDistributionData/src/generateBlocks.hpp
whaeck: Changed this in the fix/review-1 branch. Also did this in generateXSS where appropriate.

src/ACEtk/block/details/BaseDistributionData/src/generateBlocks.hpp
whaeck: This is changed in a later PR. It now says:
```
auto end = this->iterator( 2 + 2 * nr + 2 * ne + 1 );
```

src/ACEtk/block/details/BaseDistributionData/src/generateBlocks.hpp
whaeck: For my own comprehension I often do things like this.

One value for NR, followed by two times NR values, followed by the NE value. The compiler should simplify this to 2 + 2*nr easily.

src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
whaeck: Yes. Changed in fix/review-1

src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
whaeck: Changed in fix/review-1

src/ACEtk/block/details/TabulatedProbabilityDistribution/src/generateXSS.hpp
whaeck: Changed in fix/review-1

whaeck:
@cjosey fix/review-1 has been pushed and contains the changes mentioned above. In addition, I checked every generateXSS function for all blocks and made const/const auto/static_cast<double> changes as required. Seems my code is just plain sloppy :-)

python/src/block/KalbachMannDistributionData.python.cpp
whaeck: but later

python/test/block/Test_ACEtk_TabulatedOutgoingEnergyDistribution.py
whaeck: Need to revisit this at a later stage.

src/ACEtk/block/AngularDistributionData.hpp
whaeck: but later

src/ACEtk/block/KalbachMannDistributionData.hpp
whaeck: Maybe later?

src/ACEtk/block/KalbachMannDistributionData/src/ctor.hpp
whaeck: Not a bad point. Maybe later?

In most cases it is going be 
```
~KalbachMannDistributionData = default;
```

src/ACEtk/block/TabulatedKalbachMannDistribution/src/ctor.hpp
whaeck: We should look at this, but later.

src/ACEtk/block/details/BaseDistributionData.hpp
whaeck: round does return a double, so static_cast

src/ACEtk/block/details/BaseDistributionData.hpp
whaeck: Changes in fix/review-part1

src/ACEtk/block/details/BaseDistributionData/src/ctor.hpp
whaeck: maybe later

python/test/block/Test_ACEtk_KalbachMannDistributionData.py
cjosey: To verify, MCNP specifically checks if LAW 4, 44, 61 (same codepath) has discrete lines. So we can use it on our end. Whether or not the data will ever be there, who knows.

cjosey:
All components have been adequately addressed in discussions.  Here is the final `.clang-tidy` from this PR:

```
# Rationale for specific removals:
# Specific project removals:
#   We are not those projects, and these checks often have requirements
#   we do not care about.
# -llvm-header-guard
#   I like this, but the check seems to get confused by CMake and
#   tries to do the path all the way to root. I've yet to figure out why.
# -modernize-use-trailing-return-type
#   I've never seen a solid argument for the "full auto" format. It feels
#   like verbosity for the sake of verbosity
# -llvm-include-order
#   Does not match coding style
# -modernize-concat-nested-namespaces
#   Does not match coding style
# The following are of interest, but will not be added now.
# Namespace/namespace comments
# -readability-named-parameter
# -modernize-use-nodiscard
Checks: '*,-llvmlibc-*,-altera-*,-fuchsia-*,-hicpp-*,-llvm-namespace-comment,-google-readability-namespace-comments,-llvm-header-guard,-modernize-use-trailing-return-type,-llvm-include-order,-readability-named-parameter,-modernize-use-nodiscard,-modernize-concat-nested-namespaces'
# Need to use a complicated match as src/ matches [dependency]-src/
# and folders in Hana.
HeaderFilterRegex: 'src/ACEtk|src/utilities'
```

When we get near the end of this project, we can do a final pass with this. Future reviews will be done with this configuration.

cjosey (approved):

PR #47, opened 2021-10-28
Feature/dlw

whaeck:
DLW capability, currently limited to three LAWs:
- ACE LAW = 3, level scattering
- ACE LAW = 4, outgoing energy distributions
- ACE LAW = 44, Kalbach-Mann systematics

These three LAWs are between the simplest and the most complicated ones (LAW=67 is even more complicated).

The ENDF/B-VII.1 U235 ACE file uses only these LAWs so we can actually read that one's DLW block in its entirety :-)

Some abstraction has been added to remove the annoying "every law has a probability associated to it, even if there is only one". When extracting the energy distribution data from the DLW for a reaction that only has a single law (with probability = 1 everywhere), the data being returned will contain only the actual LAW data as if there was no probability associated to it. Whenever multiple LAWs are encountered, a special class that can contain multiple LAWs will be used. When constructing the DLW block, the user will just give us the LAW data if there is only one and the code will take care of adding the obligatory probability=1 data. As a result, for some LAWs we will need to give the incident energy domain where the LAW is valid (e.g. the LevelScatteringDistribution) while for other we don't have to (since the incident energy grid needs to be provided when constructing the data).

Every distribution data class will also store relative locators to the beginning of its own XSS array. When assembling the entire XSS array for the DLW block, these locators get adjusted to refer to the beginning of the DLW block, as they should.


python/src/EnergyDistributionType.python.cpp
```diff
@@ -0,0 +1,42 @@
...
+  component
+  .value( "Equiprobable",            Component::Equiprobable )
+  .value( "DiscretePhoton",          Component::DiscretePhoton )
+  .value( "LevelScattering",         Component::LevelScattering )
+  .value( "TabulatedEnergy",         Component::TabulatedEnergy )
+  .value( "GeneralEvaporation",      Component::GeneralEvaporation )
+  .value( "SimpleMaxwellianFission", Component::SimpleMaxwellianFission )
+  .value( "Evaporation",             Component::Evaporation )
+  .value( "Watt",                    Component::Watt )
+  .value( "XXXXX",                   Component::XXXXX )
```
cjosey, line 37: [Question]
Is this a placeholder for future components?

Edit: Addressed in more detail below.

python/src/block/EnergyDistributionBlock.python.cpp
```diff
@@ -0,0 +1,96 @@
...
+    "numbers in the MTR block."
+  );
+
+  // wrap the block
+  block
+  .def(
+
+    python::init< std::vector< DistributionData >&& >(),
+    python::arg( "distributions" ),
+    "Initialise the block\n\n"
```
cjosey, line 41: [Question]
Should I assume documentation to be in US English (initialize) or British English (initialise)?

python/src/block/InterpolationData.python.cpp
```diff
@@ -47,29 +47,45 @@ void wrapInterpolationData( python::module& module, python::module& ) {
...
+    "number_interpolation_regions",
+    [] ( const Block& self ) { return self.numberInterpolationRegions(); },
+    "The number of interpolation regions"
+  )
   .def_property_readonly(
 
     "NBT",
-    &Block::NBT,
+    [] ( const Block& self ) -> LongRange
+       { return self.NBT(); },
```
cjosey, line 66: [Question]
As I'm less familiar with pybind11, what does this change accomplish?

python/src/block/KalbachMannDistributionData.python.cpp
```diff
@@ -43,6 +43,64 @@ void wrapKalbachMannDistributionData( python::module& module,
     "    locb             the starting xss index with respect to the DLW block\n"
     "                     (default = 1, the relative locators get recalculated)"
   )
+  .def_property_readonly(
+
+    "LAW",
+    [] ( const Block& self ) { return self.LAW(); },
```
cjosey, line 49: [clang-tidy]
As `LAW` is a static member, it is suggesting just doing:
```cpp
[] () { return Block::LAW(); }
```
I'm not sure what this buys us (perhaps easier to optimize), so I'll note it only once.

python/src/block/OutgoingEnergyDistributionData.python.cpp
```diff
@@ -43,6 +43,64 @@ void wrapOutgoingEnergyDistributionData( python::module& module,
...
+
+    "interpolation_data",
+    [] ( const Block& self ) { return self.interpolationData(); },
+    "The interpolation data"
+  )
+  .def_property_readonly(
+
+    "NB",
+    [] ( const Block& self ) { return self.NB(); },
+    "The number of interpolation ranges"
```
cjosey, line 68: Comment differs from `number_interpolation_regions` and from other examples.

python/test/block/Test_ACEtk_EnergyDistributionBlock.py
```diff
@@ -0,0 +1,151 @@
+# standard imports
+import unittest
+
+# third party imports
+
+# local imports
+from ACEtk import EnergyDistributionBlock
+from ACEtk import TabulatedOutgoingEnergyDistribution
```
cjosey, line 8: Unused.

python/test/block/Test_ACEtk_EnergyDistributionBlock.py
```diff
@@ -0,0 +1,151 @@
...
+from ACEtk import TabulatedKalbachMannDistribution
+from ACEtk import LevelScatteringDistribution
+from ACEtk import OutgoingEnergyDistributionData
```
cjosey, line 11: Unused.

src/ACEtk/ContinuousEnergyTable.hpp
```diff
@@ -51,7 +53,7 @@ class ContinuousEnergyTable : protected Table {
     auto land = begin + this->data().JXS(8) - 1;
     auto xand = begin + this->data().JXS(9) - 1;
     auto ldlw = begin + this->data().JXS(10) - 1;
-    // auto dlw = begin + this->data().JXS(11) - 1;
+    auto dlw = begin + this->data().JXS(11) - 1;
```
cjosey, line 56: [clang-tidy - future suggestion]
For components like NXS/JXS, where their meaning is _relatively_ constant, it may be worthwhile to give these named constants as opposed to the value, such as `continuous_location_dlw` for 11. I'm not sure what would make a compact notation for it though to handle all the other table types.

src/ACEtk/DosimetryTable.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+  using Table::header;
+  using Table::data;
+  using Table::print;
+
+  //! @todo projectile() function
+  //! @todo target() function
+
+  /**
+   *  @brief Return the full ZAID or SZAID of the table
```
cjosey, line 54: [pedantic]
The formatting of the header prevents the availability of an SZAID, as it is a 10-char string -> ZZZAAA.NNp. I'm not sure if there's a clear name for an Z[A+S]ID as used here. I've been using "old SZAID" in recent work, but it's not quite well defined.

Fun fact, all our interfaces to CINDER are suffixed "AZS" due to the AAAZZZS format they use.

src/ACEtk/EnergyDistributionType.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+    Equiprobable = 1,            // ACE LAW = 1 - equiprobable bins
+    DiscretePhoton = 2,          // ACE LAW = 2 - discrete photon
+    LevelScattering = 3,         // ACE LAW = 3 - level scattering distribution
+    TabulatedEnergy = 4,         // ACE LAW = 4 - tabulated E,Eprime
+    GeneralEvaporation = 5,      // ACE LAW = 5 - general evaporation spectrum
+    SimpleMaxwellianFission = 7, // ACE LAW = 7 - simple Maxwellian fission spectrum
+    Evaporation = 9,             // ACE LAW = 9 - evaporation spectrum
+    Watt = 9,                    // ACE LAW = 11 - energy dependent Watt spectrum
+    XXXXX = 33,                  // ACE LAW = 33 - xxxx
```
cjosey, line 21: I assume once again this is a placeholder. In our code, 3 and 33 are literally the same code path with no branching, with a comment that 33 just supports more particles.

The callsite is what determines the particles emitted. If they're in DLW, they emit the projectile. If they're in DLWH, they emit something else. As far as I can tell, they never needed to be separate laws.

src/ACEtk/block/EnergyDistributionBlock/src/verifySize.hpp
```diff
@@ -0,0 +1,14 @@
+static void verifySize( Iterator begin, Iterator dlw, Iterator end,
```
cjosey, line 1: [comment]
Iterator `end` is unused in this function.

src/ACEtk/block/details/Base.hpp
```diff
@@ -107,7 +109,10 @@ class Base {
    */
   auto XSS( std::size_t index ) const {
 
-    return *( this->iterator( index ) );
+    #ifndef NDEBUG
+    this->verifyIndex( index, 1, this->length() );
+    #endif
+    return *std::next( this->begin(), index - 1 );
```
cjosey, line 115: [Design Question]
If you call `XSS(0)` which is not an unusual accidental value, this will either segfault (`NDEBUG` set, will try to get the address 2^64-1 after `begin()`) or throw (`NDEBUG` unset). Pybind11 will convert the exception.

As Python is not really designed around handling the underlying library segfaulting, I would think the Python bindings should always have `NDEBUG` unset. Thoughts?

src/ACEtk/block/details/Base/src/verifyIndex.hpp
```diff
@@ -1,11 +1,13 @@
...
+                  const std::size_t min,
+                  const std::size_t max ) const {
 
-  if ( ( index < 1 ) || ( index > this->length_ ) ) {
+  if ( ( index < min ) || ( index > max ) ) {
 
-    Log::error( "Illegal relative index argument into {} block", this->name_ );
+    Log::error( "Illegal relative index argument into {} block", this->name() );
     Log::info( "Index value: {}", index );
     Log::info( "{} may be indexed with values between 1 and {} inclusively",
```
cjosey, line 9: Should replace `1` with `{}` to take `min` as input.

cjosey (changes requested):
I think the only thing that needed changes is that one logging error message with missing `min`.

python/src/EnergyDistributionType.python.cpp
whaeck: Yes, I think this is for ACELAW=33. I didn't have a name for it back then. It eventually got named properly later.

python/src/block/InterpolationData.python.cpp
whaeck: NBT() returns a range and not all ranges are the same (any manipulation of a range creates another unique range type). As a result, we would have to bind an enormous amount of range types. The lambda converts this range into a range type called ```any_view< T >``` so that we only need to bind this particular type.

python/src/block/OutgoingEnergyDistributionData.python.cpp
whaeck: This particular one seems to have been fixed later. I've checked all files and could not find any more.

python/test/block/Test_ACEtk_EnergyDistributionBlock.py
whaeck: This has been removed in a later PR.

python/test/block/Test_ACEtk_EnergyDistributionBlock.py
whaeck: This has been removed in a later PR

src/ACEtk/block/EnergyDistributionBlock/src/verifySize.hpp
whaeck: I removed the name of the iterator to maintain the interface of the verifySize function we use in many other places.

src/ACEtk/block/details/Base/src/verifyIndex.hpp
whaeck: Good catch.

src/ACEtk/ContinuousEnergyTable.hpp
whaeck: Good idea, and it gets rid of the magic numbers. Will do this later.

cjosey (approved):

PR #48, opened 2021-10-29
Feature/dosimetry

whaeck:
Full capability for reading and creating dosimetry (type y) ACE files!

This simple ACE file type only contains the MTR and SIGD block so it is a prime candidate for testing our approach to constructing ACE files from scratch. The DosimetryTable.test.cpp contains two tests: reading an existing ACE file and creating the same file from scratch. In both tests, the IZ, AW, NXS, JXS and XSS are verified value by value. In addition, the various interface objects are tested as well to ensure that they work.

The dosimetry file used as a base is a very slightly modified version of the 13027.24y table from 531dos. The modifications made are:
- the awr value (due to a potential bug in disco)
- NXS(5) was set to zero since this value is only used when AND and DLW are present
- the XSS array had a trailing 0 value and was removed (the length values in the NXS and JXS array are adjusted accodingly)


python/examples/create-a-dosimetry-ace-file.ipynb
cjosey, line 1: I don't really have a setup for checking python notebooks, so review of these will be cursory until I get one set up.

python/test/block/Test_ACEtk_DosimetryCrossSectionData.py
```diff
@@ -0,0 +1,63 @@
...
+            self.assertEqual( 0, len( chunk.interpolants ) )
+            self.assertEqual( 0, len( chunk.NBT ) )
+            self.assertEqual( 0, len( chunk.boundaries ) )
+
+            self.assertEqual( 3, chunk.NE )
+            self.assertEqual( 3, chunk.number_energy_points )
+
+            self.assertEqual( 3, len( chunk.cross_sections ) )
+            self.assertEqual( 2., chunk.cross_sections[0] )
+            self.assertEqual( 6., chunk.cross_sections[-1] )
```
cjosey, line 42: `chunk.cross_sections` is tested twice, `chunk.energies` is not checked.

src/ACEtk/ContinuousEnergyTable/src/ctor.hpp
```diff
@@ -19,3 +19,15 @@ ContinuousEnergyTable( ContinuousEnergyTable&& cetable ) :
 
   this->generateBlocks();
 }
+
+ContinuousEnergyTable& operator=( const ContinuousEnergyTable& base ) {
+
+  new (this) ContinuousEnergyTable( base );
```
cjosey, line 25: [question]
Does this and the one below need to `this->generateBlocks()`? (This applies to other `operator=` in `ctor.hpp`)

src/ACEtk/DosimetryTable/src/generateData.hpp
```diff
@@ -0,0 +1,43 @@
...
+  jxs[6] = xss.size() + mtr.NTR() + 1;
+  xss.insert( xss.end(), sigd.begin(), sigd.end() );
+  jxs[21] = xss.size();
+
+  // set the nxs values for the dosimetry table
+  // NXS(1) = length
+  // NXS(2) = za
+  // NXS(4) = NTR
+  nxs[0] = xss.size();
+  nxs[1] = z * 1000 + a;
```
cjosey, line 38: [Future work - documentation]
It doesn't need to be here, but it may be worth documenting the historical convention on handling metastables, namely:
1. The constructor arguments for Z and A are _unmodified_ by S. Tc99m should not have A = 499.
2. The ZAID should be ZZZ[AAA + 300 (S > 0) + 100 S].NNx.
3. Am242m1 and Am242 are swapped (LANL convention - NNDC doesn't swap).

You can rightfully argue that the responsibility is on my team's side to document this. I intend to do so very soon. I just thought it would be easy for someone to have no idea about this weirdness if they're only working on this side.

src/ACEtk/DosimetryTable/test/DosimetryTable.test.cpp
```diff
@@ -0,0 +1,505 @@
...
+using Header = Table::Header;
+
+void verifyChunk( const DosimetryTable& );
+
+SCENARIO( "DosimetryTable" ){
+
+  GIVEN( "valid data for a DosimetryTable" ) {
+
+    auto table = fromFile( "13027.24y" );
+    std::array< int32_t, 16 > iz = table.data().IZ();
```
cjosey, line 23: [Design suggestion - future]
Should `iz_size`, etc, be constants, or even have `using iz_array = std::array<int32_t, 16>`?

src/ACEtk/DosimetryTable/test/DosimetryTable.test.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+  CHECK( 1 == chunk.SIGD().LSIG( 1 ) );
+  CHECK( 615 == chunk.SIGD().LSIG( 2 ) );
+  CHECK( 1 == chunk.SIGD().crossSectionLocator( 1 ) );
+  CHECK( 615 == chunk.SIGD().crossSectionLocator( 2 ) );
+
+  auto xs1 = chunk.SIGD().crossSectionData( 1 );
+  auto xs2 = chunk.SIGD().crossSectionData( 2 );
+  CHECK( 306 == xs1.numberEnergyPoints() );
+  CHECK( 272 == xs2.numberEnergyPoints() );
```
cjosey, line 500: [Suggestion]
Should this also test for energy point values?

src/ACEtk/block/DosimetryCrossSectionData/test/DosimetryCrossSectionData.test.cpp
```diff
@@ -0,0 +1,104 @@
...
+  CHECK( 0 == chunk.numberInterpolationRegions() );
+  CHECK( 0 == chunk.INT().size() );
+  CHECK( 0 == chunk.interpolants().size() );
+  CHECK( 0 == chunk.NBT().size() );
+  CHECK( 0 == chunk.boundaries().size() );
+
+  CHECK( 3 == chunk.NE() );
+  CHECK( 3 == chunk.numberEnergyPoints() );
+
+  CHECK( 3 == chunk.crossSections().size() );
```
cjosey, line 97: Double cross section test also here.

src/ACEtk/block/details/BaseCrossSectionBlock.hpp
```diff
@@ -0,0 +1,115 @@
...
+
+protected:
+
+  /* constructor */
+  #include "ACEtk/block/details/BaseCrossSectionBlock/src/ctor.hpp"
+
+  /**
+   *  @brief Return the iterator to the start of the sig block
+   */
+  Iterator sig() const { return this->sig_; }
```
cjosey, line 37: [clang-tidy]
This one's interesting. `ctor.hpp` already declared the block after it as public. As a result, the function `Iterator sig()` is not protected. I think this is the general weakness to using the `#include` fragments.

src/ACEtk/block/details/BaseTabulatedData/src/generateBlocks.hpp
```diff
@@ -0,0 +1,13 @@
+void generateBlocks() {
+
+  std::size_t nr = static_cast< std::size_t >( this->XSS( 1 ) );
+  std::size_t ne = static_cast< std::size_t >( this->XSS( 1 + 2 * nr + 1 ) );
+  auto begin = this->begin();
+  auto data = begin + 2 * nr + 1;
+  auto end = data + + 2 * ne + 1;
```
cjosey, line 7: This `+  +` hasn't yet been fixed in the latest branch.

cjosey (changes requested):
Well, two down. Only a few scattered comments. Another `+ +` issue, a few test thoroughness concerns, and some documentation notes.

python/examples/create-a-dosimetry-ace-file.ipynb
whaeck: That's fine. Most of the python notebooks are works in progress

python/test/block/Test_ACEtk_DosimetryCrossSectionData.py
whaeck: This was corrected in a later PR.

src/ACEtk/ContinuousEnergyTable/src/ctor.hpp
whaeck: I don't think so. These use the other constructors that call generateBlocks() under the hood so we should be fine.

src/ACEtk/DosimetryTable/src/generateData.hpp
whaeck: Yeah, we should probably change this. I'll make this into an issue.

src/ACEtk/block/DosimetryCrossSectionData/test/DosimetryCrossSectionData.test.cpp
whaeck: This was changed in a future PR.

src/ACEtk/block/details/BaseCrossSectionBlock.hpp
whaeck: This class was refactored later (into BasBlockWIthLocators) and this issue was resolved as well. The ctor file now declares one private constructor while all others are protected. The equivalant to the sig() function is thus also protected.

I do agree that these types of issue can pop up due to the use of include files.

cjosey (approved):

PR #49, opened 2021-11-18
Feature/dlw components 3

whaeck:
Three more energy distribution options:
- discrete photon
- equiprobable outgoing energy bins
- general evaporation spectrum


python/src/block/DiscretePhotonDistribution.python.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    "maximum_incident_energy",
+    &Block::maximumIncidentEnergy,
+    "The maximum incident energy for the distribution"
+  )
+  .def_property_readonly(
+
+    "LP",
+    &Block::LP,
+    "The primary flag value (0, 1 or 2)"
```
cjosey, line 75: [design question, future]
Should these return enums or similar instead?

python/src/block/GeneralEvaporationSpectrum.python.cpp
```diff
@@ -0,0 +1,171 @@
...
+  )
+  .def_property_readonly(
+
+    "NET",
+    &Block::NET,
+    "The number of x values"
+  )
+  .def_property_readonly(
+
+    "number_bins",
```
cjosey, line 155: [aside]
I almost got caught out by `number_bins` not being `NET` and was going to comment their documentation is different.

python/test/block/Test_ACEtk_GeneralEvaporationSpectrum.py
```diff
@@ -0,0 +1,86 @@
...
+# third party imports
+
+# local imports
+from ACEtk import EnergyDistributionType
+from ACEtk import GeneralEvaporationSpectrum
+
+class Test_ACEtk_GeneralEvaporationSpectrum( unittest.TestCase ) :
+    """Unit test for the GeneralEvaporationSpectrum class."""
+
+    chunk = [ 0,
```
cjosey, line 13: [clarification]
I'm a bit confused on the value of NR. In the ACE format document, it indicates that if NR = 0, then you use linear-linear interpolation.

Is this related to the multi-law stuff and thus I won't see any tests with a non-zero value until then? It is tricky to figure out what this does from the format manual.

src/ACEtk/block/AngularDistributionData/src/generateXSS.hpp
```diff
@@ -3,14 +3,20 @@ generateXSS( std::vector< Distribution >&& distributions, std::size_t locb ) {
 
   std::size_t ne = distributions.size();
   std::vector< double > xss( 1 + 2 * ne );
+
+  auto incidentEnergy = [] ( const auto& value )
+                           { return value.incidentEnergy(); };
+
+  std::sort( distributions.begin(), distributions.end(),
+             [=] ( const auto& left, const auto& right )
```
cjosey, line 11: I don't think the `[=]` is doing anything as all the variables I see used are passed as arguments, but if it were, would it be better to `[&]`?

src/ACEtk/block/AngularDistributionData/src/generateXSS.hpp
```diff
@@ -3,14 +3,20 @@ generateXSS( std::vector< Distribution >&& distributions, std::size_t locb ) {
...
   std::size_t index = 1;
   for ( const auto& distribution : distributions ) {
 
     // set the energy value
-    xss[index] = std::visit( [] ( const auto& value )
-                                { return value.incidentEnergy(); },
-                             distribution );
+    xss[index] = std::visit( incidentEnergy, distribution );
 
     // set the locator
```
cjosey, line 21: [non-PR aside]
Should this comment note that it is LOCC? I was baffled by it setting a negative value until I tracked that down.

src/ACEtk/block/DiscretePhotonDistribution.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   *  @brief Return the primary flag value (0, 1 or 2)
+   */
+  unsigned int LP() const { return this->XSS( 1 ); }
+
+  /**
+   *  @brief Return the primary flag value (0, 1 or 2)
+   */
+  unsigned int primaryPhotonFlag() const { return this->LP(); }
```
cjosey, line 70: [question]
LP=2 has different kinematics than LP=0 or 1 (and takes the excess kinetic energy from an incident particle, making it "primary"). What distinguishes LP=0 from LP=1?

src/ACEtk/block/EquiprobableOutgoingEnergyBinData/src/generateXSS.hpp
```diff
@@ -0,0 +1,35 @@
+static std::vector< double >
+generateXSS(
+  std::vector< long >&& boundaries,
+  std::vector< long >&& interpolants,
+  std::vector< EquiprobableOutgoingEnergyBins >&& distributions ) {
+
+  // reserve size in the vector
+  std::size_t nr = boundaries.size();
+  std::size_t ne = distributions.size();
+  std::vector< double > xss( 2 + ne ); // NE, list of energies, NET
```
cjosey, line 10: [comment]
This way of constructing XSS appears to be 100% correct. However, I had to spend more time than I'd like proving it to myself due to the nonlinearity going on (allocate, prepend, set values, append).

src/ACEtk/block/EquiprobableOutgoingEnergyBins.hpp
```diff
@@ -0,0 +1,63 @@
...
+   */
+  double incidentEnergy() const { return this->incident_; }
+
+  /**
+   *  @brief Return the number of equiprobable angular bins
+   */
+  int numberBins() const { return this->length() - 1; }
+
+  /**
+   *  @brief Return the cosine values
```
cjosey, line 44: Documentation is incorrect.

src/ACEtk/block/GeneralEvaporationSpectrum/src/ctor.hpp
```diff
@@ -0,0 +1,80 @@
...
+  std::size_t ne = static_cast< std::size_t >( this->XSS( 1 + 2 * nr + 1 ) );
+  std::size_t nb = static_cast< std::size_t >( this->XSS( 1 + 2 * nr + 1 + 2 * ne + 1 ) );
+  verifySize( this->begin(), this->end(), nr, ne, nb );
+  this->generateBlocks();
+}
+
+GeneralEvaporationSpectrum& operator=( const GeneralEvaporationSpectrum& base ) {
+
+  new (this) GeneralEvaporationSpectrum( base );
+  return *this;
```
cjosey, line 73: [question]
How do you prevent memory leaks caused by repeatedly doing:
```
GeneralEvaporationSpectrum& x = y;
```
and letting `x` go out of scope - say, in a loop?

[Edit] did some research, did not know about placement-new. In discussions, people note that since `new` is not exception-safe, this can create malformed objects on exception. Also, it may not be well-defined if you do `x = x`. Which is odd but possible.

cjosey (changes requested):
Only thing blocking approval is the incorrect doxygen comment. Rest is just questions, design stuff, and me probably getting confused by pointers.

python/src/block/DiscretePhotonDistribution.python.cpp
whaeck: Yes, we probably should do that. I'll make that into an issue.

python/test/block/Test_ACEtk_GeneralEvaporationSpectrum.py
whaeck: It seems to be an option in the ACE format but I've never seen anything that uses NR != 0. We can talk about this in more detail when we go over these things after my updates following the review.

src/ACEtk/block/AngularDistributionData/src/generateXSS.hpp
whaeck: The incidentEnergy lambda should get captured here.

src/ACEtk/block/AngularDistributionData/src/generateXSS.hpp
whaeck: I updated the comment to be more clear on what is happening

src/ACEtk/block/EquiprobableOutgoingEnergyBins.hpp
whaeck: I corrected this. I also checked the python side where it seems to be correct.

src/ACEtk/block/GeneralEvaporationSpectrum/src/ctor.hpp
whaeck: I probably should look into this some more.

cjosey (approved):

PR #51, opened 2021-11-29
Feature/dlw components 5

whaeck:
Adding the case of multi-law data.

Some refactoring of the generate xss function in the DLW block, removed && semantics from constructors to allow for both copying and moving data.


src/ACEtk/block/EnergyDistributionBlock/src/generateXSS.hpp
```diff
@@ -10,108 +10,88 @@ generateXSS( std::vector< DistributionData >&& distributions ) {
...
+
+      std::visit(
+
+        utility::overload{
+
+          [ &xss, offset ] ( const MultiDistributionData& value ) {
+
+            // remake the internal xss array with the proper locators
+            MultiDistributionData temp(
+                std::move( value.probabilities() ),
```
cjosey, line 24: [clang-tidy]
Got an interesting one here. "std::move of the const expression has no effect."
Appears in several other locations.

src/ACEtk/block/MultiDistributionData/src/ctor.hpp
```diff
@@ -0,0 +1,62 @@
...
+ *
+ *  @param[in] locb    the starting xss index with respect to the superblock
+ *  @param[in] sig     the begin iterator of the block in the XSS array
+ *  @param[in] end     the end iterator of the block in the XSS array
+ */
+MultiDistributionData( std::size_t locb, Iterator begin, Iterator end ) :
+  Base( "DLW::MultiDistributionData", begin, end ),
+  locb_( locb ) {
+
+//  std::size_t nr = static_cast< std::size_t >( this->XSS( 1 ) );
```
cjosey, line 45: [question]
Why are these commented out?

cjosey (approved):

src/ACEtk/block/EnergyDistributionBlock/src/generateXSS.hpp
whaeck: Interesting. I should change the argument of the lambda to auto&&. These things get eviscerated anyway.

src/ACEtk/block/MultiDistributionData/src/ctor.hpp
whaeck: I have no idea. In the latest version, these comments are no longer here.

src/ACEtk/block/EnergyDistributionBlock/src/generateXSS.hpp
whaeck: I have gone over all lambdas and where appropriate, changed the argument from const& to &&

whaeck: Correction: I put back the original lambdas since there seems to be an issue that causes the xss arrays to no longer be as expected

PR #53, opened 2021-12-09
Feature/photons

whaeck:
Full support for the secondary photon data in an ACE file.

@nathangibson14 Happy work anniversary. You're welcome ;-)


whaeck:
@nathangibson14 This should now fix the issue you saw with n+He3. As you will notice, the block generation is a lot more intelligent now. Instead of just using the locators of the next block as the end point of the current one, the code now searches for the first locator value following the current one that is equal to or larger than the current one (if it is equal, the block is empty). Following this change, I also made a few more fixes (in size checks, e.g. for the YP block).

I also added a second CE table test using the He3 file.

python/src/block/PhotonProductionCrossSectionData.python.cpp
```diff
@@ -0,0 +1,80 @@
...
+    "Arguments:\n"
+    "    self      the block\n"
+    "    index     the energy index\n"
+    "    values    the cross section values"
+  )
+  .def_property_readonly(
+
+    "MFTYPE",
+    &Block::MFTYPE,
+    "The representation type (should always be 12 or 16))"
```
cjosey, line 48: Is this necessarily true? The ACE Format indicates that MFTYPE can be 13 if it's in the SIGP block.

Also, double-close-parenthesis.

python/src/block/TabulatedFissionMultiplicity.python.cpp
```diff
@@ -32,6 +32,30 @@ void wrapTabulatedFissionMultiplicity( python::module& module, python::module& )
...
+                  std::vector< double >, std::vector< double > >(),
+    python::arg( "boundaries" ), python::arg( "interpolants" ),
+    python::arg( "energies" ), python::arg( "xs" ),
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self            the block\n"
+    "    boundaries      the interpolation range boundaries\n"
+    "    interpolants    the interpolation types for each range\n"
+    "    energies        the energy values\n"
+    "    multiplicities   the multiplicity values"
```
cjosey, line 57: Documentation and argument name not in sync.

python/src/block/TabulatedSecondaryParticleMultiplicity.python.cpp
```diff
@@ -0,0 +1,183 @@
...
+                  std::vector< long >, std::vector< long >,
+                  std::vector< double >, std::vector< double > >(),
+    python::arg( "mftype" ), python::arg( "mt" ),
+    python::arg( "boundaries" ), python::arg( "interpolants" ),
+    python::arg( "energies" ), python::arg( "xs" ),
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self            the block\n"
+    "    mftype           the type (12 or 16)\n"
+    "    mt               the associated reaction number\n"
```
cjosey, line 62: I don't know if this is something you want to spend time on, but misaligned comment block.

src/ACEtk/block/AngularDistributionBlock.hpp
```diff
@@ -76,9 +78,9 @@ class AngularDistributionBlock : protected details::Base {
   int LAND( std::size_t index ) const {
 
     #ifndef NDEBUG
-    this->verifyReactionIndex( index, 0, this->NR() );
+    this->verifyReactionIndex( index, this->offset_ == 1 ? 0 : 1, this->NR() );
```
cjosey, line 81: This is fine and the correct way to do this, but my brain wants to do `!(this->offset_ == 1)` since it habitually avoids ternary operators.

src/ACEtk/block/AngularDistributionBlock/src/insertElastic.hpp
```diff
@@ -0,0 +1,7 @@
+std::vector< DistributionData >
+insertElastic( DistributionData&& elastic,
```
cjosey, line 2: [clang-tidy]
This member function can be made static.

src/ACEtk/block/MultiplicityReactionNumberBlock.hpp
```diff
@@ -0,0 +1,137 @@
...
+   *  @param[in] reaction    the reaction number
+   */
+  std::size_t index( unsigned int reaction ) const {
+
+    auto iter = std::find( this->begin() + 1, this->end(), reaction );
+    if ( iter != this->end() ) {
+
+      return std::distance( this->begin() + 1, iter ) + 1;
+    }
+    else {
```
cjosey, line 116: There's a habit I've picked up that if an `if` always returns, I don't write the `else`. I think though it only helps in readability on blocks that are longer than this.

src/ACEtk/block/MultiplicityReactionNumberBlock/src/ctor.hpp
```diff
@@ -0,0 +1,38 @@
+private:
+
+/**
+ *  @brief Private constructor
+ */
+MultiplicityReactionNumberBlock( std::vector< unsigned int >&& reactions,
+                                 std::size_t nyp ) :
```
cjosey, line 7: `nyp` is unused.

src/ACEtk/block/MultiplicityReactionNumberBlock/src/verifySize.hpp
```diff
@@ -0,0 +1,14 @@
+static void verifySize( Iterator begin, Iterator end, unsigned int nyp ) {
+
+  if ( nyp != 0 ) {
+
+    auto size = std::distance( begin, end );
+    if ( size < nyp + 1 ) {
```
cjosey, line 6: [design question]
Is there motivation for these checks to not be `!=`?

cjosey (commented):
Only a couple of minor comments.

python/src/block/PhotonProductionCrossSectionData.python.cpp
whaeck: This is indeed wrong. MFTYPE is always 13 in this case. I changed the documentation accordingly.

python/src/block/TabulatedFissionMultiplicity.python.cpp
whaeck: Indeed. I corrected it.

python/src/block/TabulatedSecondaryParticleMultiplicity.python.cpp
whaeck: I tend to like nicely indented code and documentation, so yes: if you see any of these you can either tell me or just correct them in a separate PR (for the ones you pointed out, I have done it myself as I do not mind making these corrections) :-)

src/ACEtk/block/AngularDistributionBlock.hpp
whaeck: You have a point here. I do think the ternary operator adds readability though.

src/ACEtk/block/AngularDistributionBlock/src/insertElastic.hpp
whaeck: This was refactored into a base angular distribution block. It still had the error, I corrected it.

src/ACEtk/block/MultiplicityReactionNumberBlock.hpp
whaeck: Yeah, I can see that. I don't really have a strong opinion here.

src/ACEtk/block/MultiplicityReactionNumberBlock/src/ctor.hpp
whaeck: Seems I refactored and did not remove this useless piece of code. I removed it.

src/ACEtk/block/MultiplicityReactionNumberBlock/src/verifySize.hpp
whaeck: It is technically not forbidden to have "gaps" at the end of a block (i.e. junk data between the end of the current block and the next block). I have used this approach to correct data in an ACE file without changing the locators in NJOY2016.

As such, all verifySize functions verify that the size is AT LEAST the minimum required size. Any size larger than the minimum required size is accepted.

cjosey (approved):

PR #54, opened 2022-01-05
Feature/refactor

whaeck:
Introduces base classes for blocks with locators (cross section and energy distribution blocks) and the angular distribution blocks.

The different blocks now allow for extracting vectors of subblocks whenever they are needed (e.g. all cross section data objects from the SIG block).

More tests.

@nathangibson14 Happy New Year ;-)


python/src/block/SecondaryParticleAngularDistributionBlock.python.cpp
```diff
@@ -0,0 +1,126 @@
...
+  // wrap views created by this block
+
+  // create the block
+  python::class_< Block > block(
+
+    module,
+    "SecondaryParticleAngularDistributionBlock",
+    "The SecondaryParticleAngularDistributionBlock class contains angular\n"
+    "distribution data, one for each the NXS(6) reaction numbers on the MTRP\n"
+    "block (for the LANDP,ANDP blocks) and each of the NTRO(ITYPE) reactions on\n"
```
cjosey, line 30: [low priority]
Space between LANDP, ANDP.

python/src/block/SecondaryParticleAngularDistributionBlock.python.cpp
```diff
@@ -0,0 +1,126 @@
...
+    "(debug mode only).\n\n"
+    "    self     the block\n"
+    "    index    the index (one-based, zero for elastic)"
+  )
+  .def(
+
+    "is_given",
+    &Block::isGiven,
+    python::arg( "index" ),
+    "Return whether or not the distribution data is fully isotropic for\n"
```
cjosey, line 97: This documentation seems incorrect.

src/ACEtk/block/AngularDistributionBlock/test/AngularDistributionBlock.test.cpp
```diff
@@ -11,17 +11,17 @@ using AngularDistributionBlock = block::AngularDistributionBlock;
...
 using DistributionGivenElsewhere = block::DistributionGivenElsewhere;
-using DistributionData = AngularDistributionBlock::DistributionData;
+using DistributionData = block::AngularDistributionBlock::DistributionData;
 using TabulatedAngularDistribution = block::TabulatedAngularDistribution;
 using IsotropicAngularDistribution = block::IsotropicAngularDistribution;
 
 std::vector< double > chunk();
 void verifyChunk( const AngularDistributionBlock& );
-void verifyChunkWithoutElastic( const AngularDistributionBlock& );
+//void verifyChunkWithoutElastic( const AngularDistributionBlock& );
```
cjosey, line 20: I know it's just a test, but you might want to clean this up.

src/ACEtk/block/MultiDistributionData/src/ctor.hpp
```diff
@@ -1,15 +1,29 @@
...
+    this->probabilities_.clear();
+    this->distributions_.clear();
+    this->generateBlocks();
+  }
 }
 
 MultiDistributionData( MultiDistributionData&& base ) :
-  Base( std::move( base ) ) {
+  Base( std::move( base ) ), locb_( base.locb_ ),
+  probabilities_( std::move( base.probabilities_ ) ),
```
cjosey, line 18: Is this safe? `base` has been `move`d, so `base.probabilities_` may not necessarily exist.

src/ACEtk/block/SecondaryParticleAngularDistributionBlock.hpp
```diff
@@ -0,0 +1,152 @@
...
+  using DistributionData = details::BaseAngularDistributionBlock::DistributionData;
+
+  /* constructor */
+  #include "ACEtk/block/SecondaryParticleAngularDistributionBlock/src/ctor.hpp"
+
+  /* methods */
+
+  /**
+   *  @brief Return the number of reactions that produce the projectile
+   *         (excluding elastic )
```
cjosey, line 45: Does elastic make sense in this block?

src/ACEtk/block/SecondaryParticleAngularDistributionBlock/src/ctor.hpp
```diff
@@ -0,0 +1,29 @@
+SecondaryParticleAngularDistributionBlock() = default;
+
+SecondaryParticleAngularDistributionBlock( const SecondaryParticleAngularDistributionBlock& ) = default;
+
+SecondaryParticleAngularDistributionBlock( SecondaryParticleAngularDistributionBlock&& ) = default;
+
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] data    the ditribution data for all reactions
```
cjosey, line 10: "ditribution"

src/ACEtk/block/details/BaseAngularDistributionBlock/src/insertZero.hpp
```diff
@@ -0,0 +1,6 @@
+std::vector< DistributionData >
+insertZero( DistributionData&& zero, std::vector< DistributionData >&& data ) {
```
cjosey, line 2: This function can be made static.

src/ACEtk/block/details/BaseBlockWithLocators/src/ctor.hpp
```diff
@@ -0,0 +1,81 @@
...
+  if ( Base::owner() ) {
+
+    this->iterator_ = this->begin() + this->nr_;
+    this->data_.clear();
+    this->generateBlocks();
+  }
+}
+
+BaseBlockWithLocators( BaseBlockWithLocators&& base ) :
+  Base( std::move( base ) ), nr_( base.nr_ ), iterator_( base.iterator_ ),
```
cjosey, line 32: Same question as earlier, but this one has `base` use after move.

Edit: This general form appears in many locations, so if changed, you should probably reassess most of them.

cjosey (changes requested):
Biggest concern is the many use-after-move's. I'm not sure if any of them pose an issue, but it is worth checking. The rest is really just documentation.

python/src/block/SecondaryParticleAngularDistributionBlock.python.cpp
whaeck: This one might be a bit obscure. When the angular distribution is fully isotropic, no data is given.

src/ACEtk/block/SecondaryParticleAngularDistributionBlock.hpp
whaeck: It absolutely does not for incident neutron and incident charged particle ACE files. And in photonuclear it makes no sense to have an elastic so not there either? Probably a leftover from copy/pasting since this block looks like the AND one.

whaeck: It seems this has been corrected already.

src/ACEtk/block/details/BaseAngularDistributionBlock/src/insertZero.hpp
whaeck: This has been corrected already in another PR

src/ACEtk/block/MultiDistributionData/src/ctor.hpp
whaeck: As I understand it, the MultiDistributionData move constructor casts ```base``` to an rvalue using std::move, then passes the result to the Base move constructor, which involves an implicit cast from MultiDistributionData&& to Base&&. The Base move constructor might steal the guts of the base class members of ```base```, but it can't touch the guts of the derived class members because it only sees a Base&&.

I therefore think this is safe - as long as I do not need base class members from that point on.

python/src/block/SecondaryParticleAngularDistributionBlock.python.cpp
cjosey: To clarify, it seems like the first line was copy-pasted from the block above, as it currently reads "Return whether or not the distribution data is fully isotropic for is given in the AND block" as compared to the C++ side "Return whether or not the distribution data for a reaction index is given in the AND block"

whaeck: yeah, i saw that after i commented. I chganged it accordingly.

cjosey (approved):

PR #56, opened 2022-01-06
Feature/delayed

whaeck:
Work on support for delayed neutron data:
- adding the DNU block for the delayed neutron nubar
- adding the DNED block for the energy distributions
- implementing the precursor data and the precursor data block


python/src/block/DelayedNeutronPrecursorData.python.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+  // wrap views created by this block
+
+  // create the block
+  python::class_< Block > block(
+
+    module,
+    "DelayedNeutronPrecursorData",
+    "Delayed neutron precursor group data\n\n"
+    "The DelayedNeutronPrecursorData class contains the prrcursor group's decay\n"
```
cjosey, line 28: "prrcursor"

python/src/block/DelayedNeutronPrecursorData.python.cpp
```diff
@@ -0,0 +1,166 @@
...
+    python::arg( "boundaries" ), python::arg( "interpolants" ),
+    python::arg( "energies" ), python::arg( "xs" ),
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self             the block\n"
+    "    constant         the precursor group's decay constant\n"
+    "    boundaries       the interpolation range boundaries\n"
+    "    interpolants     the interpolation types for each range\n"
+    "    energies         the energy values\n"
+    "    probabilities    the probability values"
```
cjosey, line 61: `probabilities` passed to arg `xs`.

src/ACEtk/ContinuousEnergyTable.hpp
```diff
@@ -48,6 +49,10 @@ class ContinuousEnergyTable : protected Table {
   block::DLW dlwp_;
   block::YP yp_;
 
+  block::NU dnu_;
+  block::BDD bdd_;
+  block::DLW dned_;
```
cjosey, line 54: [design question]
I know it probably doesn't matter, but all of these are available in all tables. This should only be available for projectiles that can cause fission? Or would the check be "checking for the sake of checking"?

src/ACEtk/block/DelayedNeutronPrecursorBlock/src/ctor.hpp
```diff
@@ -0,0 +1,71 @@
...
+  if ( Base::owner() ) {
+
+    this->data_.clear();
+    this->generateBlocks();
+  }
+}
+
+DelayedNeutronPrecursorBlock( DelayedNeutronPrecursorBlock&& base ) :
+  Base( std::move( base ) ), npcr_( base.npcr_ ),
+  data_( std::move( base.data_ ) ) {
```
cjosey, line 31: This part should be addressed however it is addressed in the previous PR.

cjosey (approved):
All looks good here. The use-after-move follows from decisions made in the previous PR.

python/src/block/DelayedNeutronPrecursorData.python.cpp
whaeck: Yes, it should be probabilities. Tests do not pick this up because this is another one of those ctors with interpolation data.

src/ACEtk/ContinuousEnergyTable.hpp
whaeck: These have been replaced with std::optional versions later on. Relying on empty() was just too unreliable. That change also resolves the point you make.

PR #57, opened 2022-01-10
Feature/secondaries

whaeck:
The secondary particle data blocks are now enabled in the ContinuousEnergyTable:
- implemented the HPD and SIGH blocks
- enabled the TYRH, MTRH, ANDH, DLWH and YH blocks

Tests are added on the He3 case since the U235 case does not have any.


python/src/block/SecondaryParticleInformationBlock.python.cpp
```diff
@@ -0,0 +1,96 @@
...
+  // create the block
+  python::class_< Block > block(
+
+    module,
+    "SecondaryParticleInformationBlock",
+    "The continuous energy NTRO block with the number of reactions for each "
+    "secondary particle type\n\n"
+    "The SecondaryParticleInformationBlock class contains the number of reactions\n"
+    "that produce each secondary particle type as given on the PTYPE.\n"
+    "The number of available available particle types is stored in NXS(7)."
```
cjosey, line 31: "available available"

python/src/block/SecondaryParticleInformationBlock.python.cpp
```diff
@@ -0,0 +1,96 @@
...
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self       the block\n"
+    "    numbers    the number of reaction for each particle type"
+  )
+  .def_property_readonly(
+
+    "NTYPE",
+    &Block::NTYPE,
+    "The number of secondary particle types (excluding the projectile)"
```
cjosey, line 49: Is this excluding projectile, or excluding photon and projectile?

python/src/block/SecondaryParticleLocatorBlock.python.cpp
```diff
@@ -0,0 +1,183 @@
...
+  // wrap the block
+  block
+  .def(
+
+    python::init< std::vector< std::array< unsigned int, 10 > > >(),
+    python::arg( "locators" ),
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self       the block\n"
+    "    numbers    the locators for each particle type"
```
cjosey, line 42: Numbers vs. locators

python/src/block/SecondaryParticleLocatorBlock.python.cpp
```diff
@@ -0,0 +1,183 @@
...
+    "Arguments:\n"
+    "    self     the block\n"
+    "    index    the secondary particle index (one-based)"
+  )
+  .def(
+
+    "ANDH",
+    &Block::ANDH,
+    python::arg( "index" ),
+    "The locator to the ANDH block for a secondary particle index\n\n"
```
cjosey, line 143: I'll check this later, but is this not returning the block and not the locator? (this question applies to all non-L-prefixed variable documentation)

Edit: never mind. This is the locator to the ANDH block. and LANDH function call is the locator of the locator to the contents of the ANDH block.

python/test/block/Test_ACEtk_SecondaryParticleProductionBlock.py
```diff
@@ -0,0 +1,52 @@
...
+            xss = chunk.xss_array
+            for index in range( chunk.length ) :
+
+                self.assertAlmostEqual( self.chunk[index], xss[index] )
+
+        # the data is given explicitly
+        chunk = SecondaryParticleProductionBlock(
+                    index = 1,
+                    production = [ 1.71740100000E+01, 1.21438600000E+01, 2.72235400000E+02 ],
+                    heating = [ 3.63894900000E-05, 3.67662300000E-05, 3.70168600000E-05 ] )
```
cjosey, line 46: [design question]
This input allows for production and heating to be different sizes, yet there is only one length counter in the XSS for both. Is this verified anywhere? How does the code react when they are different sizes?

src/ACEtk/block/SecondaryParticleLocatorBlock.hpp
```diff
@@ -0,0 +1,189 @@
...
+
+  /**
+   *  @brief Return the locator for a secondary particle index and block index
+   *
+   *  @param[in] particle    the secondary particle index (one-based)
+   *  @param[in] index       the block index (one-based)
+   */
+  auto locator( std::size_t particle, std::size_t index ) const {
+
+    return this->XSS( ( particle - 1 ) * 10 + index );
```
cjosey, line 72: In most of the rest of the codebase, when things are simply renamed, they call another function to have a single point of truth. 

Is there a reason `locator` doesn't just call `LLOC`?

src/ACEtk/block/SecondaryParticleLocatorBlock/src/ctor.hpp
```diff
@@ -0,0 +1,42 @@
...
+SecondaryParticleLocatorBlock( SecondaryParticleLocatorBlock&& ) = default;
+
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] locators    the locators for each secondary particle type
+ */
+SecondaryParticleLocatorBlock(
+    std::vector< std::array< unsigned int, 10 > > locators ) :
+  SecondaryParticleLocatorBlock( std::move( locators ), locators.size() ) {}
```
cjosey, line 26: Use-after-move (I think)
I'm going to pause checking these until a way forward is decided, because there's quite a few of them.

src/ACEtk/block/SecondaryParticleProductionBlock/src/generateXSS.hpp
```diff
@@ -0,0 +1,24 @@
+static std::vector< double > generateXSS(
+    std::size_t index,
+    std::vector< double >&& production,
+    std::vector< double >&& heating ) {
+
+  std::size_t size = production.size();
+  if ( heating.size() != size ) {
```
cjosey, line 7: Interesting that there is a specific check for size here in this block. This does not occur elsewhere. Is there a reason? If so, is it better to check it in the `verifySize.hpp` check once?

src/ACEtk/block/SecondaryParticleProductionCrossSectionBlock.hpp
```diff
@@ -0,0 +1,117 @@
...
+namespace ACEtk {
+namespace block {
+
+/**
+ *  @class
+ *  @brief The continuous energy LSIGH and SIGH block with the secondary
+ *         particle production cross section data (multiplicity data only)
+ *
+ *  The SecondaryParticleProductionCrossSectionBlock class contains NTRO.NP sets
+ *  of secodnary particle production data, one for each reaction number on the
```
cjosey, line 21: "secodnary"

cjosey (changes requested):
A bunch of minor comments once again.

One additional question that might have been in a previous PR but I'm just noticing now. In the `PTYPE` block, are you checking that the `PTYPE` values are themselves valid? e.g., 1, 9, 31-34?

whaeck:
> One additional question that might have been in a previous PR but I'm just noticing now. In the `PTYPE` block, are you checking that the `PTYPE` values are themselves valid? e.g., 1, 9, 31-34?

No we don't check that but we probably should.

python/src/block/SecondaryParticleInformationBlock.python.cpp
whaeck: Since this can only appear in the CE table, this would be excluding the projectile and photons. The other blocks can however appear in both so it depends on the table type: for CE tables it'd be excluding the projectile and photons while for photonuclear it would be excluding the projectile only.

python/src/block/SecondaryParticleLocatorBlock.python.cpp
whaeck: These are the locators to the specific block and particle type. The IXS block only knows the locators and can therefore not return the actual blocks.

python/test/block/Test_ACEtk_SecondaryParticleProductionBlock.py
whaeck: This is checked in generateXSS(), and if they are different sizes an exception will be thrown

src/ACEtk/block/SecondaryParticleLocatorBlock.hpp
whaeck: Yup, that should have been a call to the previous one. I seem to have fixed that at a later point in time.

src/ACEtk/block/SecondaryParticleProductionBlock/src/generateXSS.hpp
whaeck: verifySize is only used for checking the iterator based constructor - which we use to construct from an existing XSS array. generateXSS is only called when constructing from scratch.

cjosey (approved):

PR #60, opened 2022-03-02
Feature/cetable

whaeck:
This pull requests adds the following:
- Extending the from-scratch constructor: adding secondary photon blocks, ptables, delayed neutron data and the remaining secondary particle blocks.
- Added tests using the U235 and He3 file to stitch an ACE file together again.
- correcting the LAW=61 and LAW=67 secondary particle distributions since ACE cannot be consistent between its' own components

The only remaining item to finish the implementation of the CE table are the energy dependent yields that can appear in the DLW block (with the locators to them actually in the TYR block). I'll be adding that in a separate pull request since I will have to do some significant rewrite there as well.


src/ACEtk/ContinuousEnergyTable/src/generateData.hpp
```diff
@@ -28,21 +57,120 @@ Data generateData( unsigned int z, unsigned int a,
     Log::info( "SIG NTR value = {}", sig.NTR() );
     throw std::exception();
   }
-  if ( ( nr != ang.NR() ) || ( ntr != dlw.NR() ) ) {
+  if ( ( nr != ang.NR() ) || ( nr != dlw.NR() ) ) {
```
cjosey, line 60: Is there any way for `nr` to not be `ang.NR()`, when it is set with it?

cjosey (approved):

PR #62, opened 2022-04-20
Feature/tsl

whaeck:
Full support for the thermal scattering ACE file:
- blocks for cross section data (inelastic and elastic) are now handled
- blocks for inelastic and elastic angular data (coherent and incoherent) are now handled
- tests for all possibilities (combinations of inalastic and coherent/incoherent elastic - using ENDF/B-VIII.0 ACE files and the experimental ENDF/B-VIII.1 D-LiD ACE file)

Some abstraction has been performed: while incoherent only gets stored in the same spot as where coherent elastic would be stored in the case of coherent only or mixed mode, there is a method to explciitly extract the coherent and incoherent - regardlss of where it is stored.

In addition, the IFENG=0/1 or 2 options have been abstracted away as well (IFENG=2 uses locators while IFENG=0/1 does not).


python/src/ThermalScatteringTable.python.cpp
```diff
@@ -0,0 +1,283 @@
...
+    "Arguments:\n"
+    "    self      the table\n"
+    "    header    the header for the table\n"
+    "    za        the list of ZA values\n"
+    "    itie      the inelastic thermal scattering cross section block\n"
+    "    itxe      the inelastic thermal scattering angular distribution block\n"
+    "    itce      the coherent elastic thermal scattering cross section block\n"
+    "    itca      the coherent elastic thermal scattering angular distribution block\n"
+    "    itcei     the incoherent elastic thermal scattering cross section block\n"
+    "    itcai     the incoherent elastic thermal scattering angular distribution block"
```
cjosey, line 60: So, I appreciate the simplification that ITCEI only points to incoherent elastic and ITCE only points to coherent elastic. But this does not follow what is in the ACE format. Is there any concern for confusion?

python/test/block/Test_ACEtk_ThermalScatteringCrossSectionBlock.py
```diff
@@ -0,0 +1,51 @@
...
+            self.assertEqual( False, chunk.empty )
+            self.assertEqual( 9, chunk.length )
+            self.assertEqual( "ThermalScatteringCrossSectionBlock", chunk.name )
+
+            self.assertEqual( 4, chunk.NE )
+            self.assertEqual( 4, chunk.number_incident_energies )
+
+            self.assertEqual( 4, len( chunk.energies ) )
+            self.assertAlmostEqual( 10., chunk.energies[0] )
+            self.assertAlmostEqual( 40., chunk.energies[-1] )
```
cjosey, line 30: [aside]
This is a very high energy S(a,b) table. 😄

src/ACEtk/ContinuousEnergyTable/src/generateBlocks.hpp
```diff
@@ -19,20 +19,30 @@ auto block( std::size_t index ) const {
         // JXS(22) is not a locator, it points to the last value of the conventional
         // XSS array (i.e. before the secondary particle data)
         end = *iter != this->data().JXS(22) ? *iter : *iter + 1;
+
+        // JXS(21) does not point to an actual block so set it to END + 1 instead
```
cjosey, line 23: Why are we adding the FIS block here? Is this just assorted fixes that would normally be in another PR?

Also, fun fact, MCNP always reconstructs the FIS block to simplify coding, mainly so it doesn't have to loop over MTR to find 18.

src/ACEtk/ThermalScatteringTable/src/generateData.hpp
```diff
@@ -0,0 +1,123 @@
...
+    throw std::exception();
+  }
+
+  // generate the xss array and set the locators in the jxs array as we go
+  jxs[0] = 1;
+  jxs[1] = jxs[0] + ne + 1;
+  xss.insert( xss.end(), itie.begin(), itie.end() );
+  jxs[2] = xss.size() + 1;
+  xss.insert( xss.end(), itxe.begin(), itxe.end() );
+  if ( idpnc ) {
```
cjosey, line 70: If statement is irrelevant.

src/ACEtk/ThermalScatteringTable/src/generateData.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+    if ( idpnc == 3 ) {
```
cjosey, line 72: I imagine this block could be cleaned up by just setting the destination indices in the 30-52 lines and going:
```cpp
if (itce.has_value()) {
  jxs[coherent_elastic_ind] = ...;
  jxs[coherent_elastic_ind+1] = ...;
}
if (itcei.has_value()) {
  jxs[incoherent_elastic_ind] = ...;
  jxs[incoherent_elastic_ind+1] = ...;
}
```
Noting that incoherent elastic is always after coherent.

src/ACEtk/block/ThermalScatteringDiscreteCosinesWithProbability/src/ctor.hpp
```diff
@@ -0,0 +1,39 @@
+ThermalScatteringDiscreteCosinesWithProbability() = default;
+
+ThermalScatteringDiscreteCosinesWithProbability( const ThermalScatteringDiscreteCosinesWithProbability& ) = default;
+ThermalScatteringDiscreteCosinesWithProbability( ThermalScatteringDiscreteCosinesWithProbability&& ) = default;
+
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] energy     the outgoing energy value
+ *  @param[in] cosines    the discrete cosine values
```
cjosey, line 10: Documentation missing `pdf`, `cdf`.

cjosey (approved):

python/src/ThermalScatteringTable.python.cpp
whaeck: The interface makes it look like that, but in reality we still have the following locators in the actual ACE file:
- ITCE points to the incoherent elastic and ITCEI is 0 if IDPNC=3
- ITCE points to the coherent elastic and ITCEI is 0 if IDPNC=4
- ITCE points to the coherent elastic and ITCEI points to incoherent elastic if IDPNC=5

The interface only abstracts away the IDPNC shenanigans while leaving these in the ACE file.

src/ACEtk/ContinuousEnergyTable/src/generateBlocks.hpp
whaeck: Most likely something I had to fix while I was working on the thermal scattering. I don;t remember.

If the FIS block is not really needed, we can deprecate it?

src/ACEtk/ThermalScatteringTable/src/generateData.hpp
whaeck: Consider it corrected.

src/ACEtk/block/ThermalScatteringDiscreteCosinesWithProbability/src/ctor.hpp
whaeck: Consider it changed.

whaeck:
@cjosey I put the changes in the PR for the previous review - since the number of changes is quite limited.

src/ACEtk/ContinuousEnergyTable/src/generateBlocks.hpp
cjosey: Ah, I more meant, the coding in MCNP always uses `FIS` for transport, and if it isn't present it reconstructs it. So we very much need it but it isn't required NJOY generates it. It just makes initialization faster if `FIS` is present.

whaeck: Ah, I see.

PR #63, opened 2022-05-11
Feature/multi ace

whaeck:
Reading concatenated ACE files


whaeck:
I think all this is in feature/tsl anyway

PR #64, opened 2022-06-06
Feature/photoatomic

whaeck:
Initial support for the photoatomic ACE file for the main blocks that are always present:
- the principal data in the ESZG block and the heating numbers
- the incoherent scattering functions and the coherent form factors


src/ACEtk/block/details/ArrayData/src/ctor.hpp
```diff
@@ -0,0 +1,95 @@
...
+ */
+template < typename Number >
+ArrayData( std::string&& name,
+           std::vector< Number >&& array1,
+           std::vector< Number >&& array2,
+           std::vector< Number >&& array3,
+           std::vector< Number >&& array4,
+           std::vector< Number >&& array5 ) :
+  ArrayData( std::move( name ),
+             std::vector< std::vector< Number > >{
```
cjosey, line 45: I see these constructors are only used rarely, but wouldn't it be easier to implement this and the 6-array versions as a variadic template?  It looks like forwarding the variadic arguments to the vector constructor is a bit complicated though...

src/ACEtk/PhotoAtomicTable.hpp
```diff
@@ -0,0 +1,147 @@
...
+  unsigned int numberEnergyPoints() const { return this->NES(); }
+
+  // JXS information
+
+  // XSS blocks
+
+  /**
+   *  @brief Return the principal cross section block
+   */
+  const block::ESZG& ESZG() const { return this->eszg_; }
```
cjosey: Is there an updated version of the ACE format manual with photoatomic data added? The one I reference just has the NXS and JXS values.

(Review paused here due to end of day)

src/ACEtk/block/details/ArrayData/src/ctor.hpp
whaeck: I actually thought of that, but didn't want to put in the time to make it work since it is a base class and as such an implementation detail that no one will interact with directly.

src/ACEtk/PhotoAtomicTable.hpp
whaeck: The ACE format manual is not up to date. You should use the MCNP5 manual appendix instead.

python/src/block/PhotoAtomicPrincipalCrossSectionBlock.python.cpp
```diff
@@ -0,0 +1,109 @@
...
+  // wrap views created by this block
+
+  // create the block
+  python::class_< Block > block(
+
+    module,
+    "PhotoAtomicPrincipalCrossSectionBlock",
+    "The photoatomic ESZG block with the principal cross sections\n\n"
+    "The PhotoAtomicPrincipalCrossSectionBlock class contains 5 arrays of the\n"
+    "same length (all values are stored as the natural logarithm):\n"
```
cjosey: Should there be any notes about how precisely zero is not exp(0) = 1, but just zero?

(man I hate how many badly-designed patches we have from over the decades -- why couldn't it just be "any value <-1000 is zero"?)

src/ACEtk/PhotoAtomicTable.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   *  @brief Return the atomic weight ratio (with respect to the neutron mass)
+   */
+  double atomicWeightRatio() const { return this->header().atomicWeightRatio(); }
+
+  /**
+   *  @brief Return the temperature of the table
+   */
+  double TEMP() const { return this->header().TEMP(); }
```
cjosey: [question]
Is there any libraries in which photoatomic temperature is not zero?  Also, I am now suspicious that the bound electron broadening (that MCNP already does) will swamp out any effects even if the data was broadened for target nucleus motion.

src/ACEtk/PhotoAtomicTable/src/ctor.hpp
```diff
@@ -0,0 +1,84 @@
...
+ *  @param[in] table    the thermal scattering table to be copied
+ */
+PhotoAtomicTable( const PhotoAtomicTable& table ) :
+  Table( table ) {
+
+  this->generateBlocks();
+}
+
+/**
+ *  @brief Move constructor using a thermal scattering table
```
cjosey: Is this a copy-paste error (same with above)?

src/ACEtk/PhotoAtomicTable/src/generateData.hpp
```diff
@@ -0,0 +1,61 @@
...
+  jxs[1] = xss.size() + 1;
+  xss.insert( xss.end(), jinc.begin(), jinc.end() );
+  jxs[2] = xss.size() + 1;
+  xss.insert( xss.end(), jcoh.begin(), jcoh.end() );
+  jxs[4] = xss.size() + 1;
+  xss.insert( xss.end(), lhnm.begin(), lhnm.end() );
+
+  // set the nxs values for the continuous energy table
+  // NXS(1) = length
+  // NXS(3) = NE
```
cjosey: NXS(2) needs to be Z for all versions of MCNP prior to 6.3.1, and there's not yet really a way in this setup to get Z from the ZAID.

src/ACEtk/PhotoAtomicTable/src/generateData.hpp
```diff
@@ -0,0 +1,61 @@
...
+  aw.fill( 0. );
+  nxs.fill( 0 );
+  jxs.fill( 0 );
+  std::vector< double > xss;
+
+  // assign za values
+  for ( unsigned int i = 0; i < za.size(); ++i ) {
+
+    iz[i] = za[i];
+    aw[i] = awr[i];
```
cjosey: Does the photoatomic data need the IZ array?

Also, the size of `za` is compared to `awr` after they are written, so if `za` is bigger than `awr`, memory errors can occur before being checked.

src/ACEtk/block/CoherentFormFactorBlock.hpp
```diff
@@ -0,0 +1,83 @@
...
+   *  @brief Return the number of values
+   */
+  constexpr unsigned int numberValues() const { return this->NM(); }
+
+  /**
+   *  @brief Return the electron recoil momentum values
+   */
+  constexpr std::array< double, 55 > momentum() const {
+
+    return {{ 0., .01, .02, .03, .04, .05, .06, .08, .10, .12, .15,
```
cjosey, line 52: If you want EPRData support, basically all you need to do (here) is make this array an optional part of the constructor. If provided, then write it before the integrated array. MCNP currently identifies the new format by checking JXS differences and if they're not (here) 2*55, it assumes there's 3 arrays and not 2. Which is a bit silly, but that's another matter.

Extended form factors are not specific to EPRData. The code converts to this format no matter what. Though I'd have to check if it requires both this and the other momentum block are simultaneously there or not.

src/ACEtk/block/IncoherentScatteringFunctionBlock.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+  /* constructor */
+  #include "ACEtk/block/IncoherentScatteringFunctionBlock/src/ctor.hpp"
+
+  /* methods */
+
+  /**
+   *  @brief Return the number of values
+   */
+  constexpr unsigned int NM() const { return 21; }
```
cjosey: Same here, if the difference in JXS is != 21, it assumes there's two arrays with the first containing momentum.

cjosey (dismissed):

python/src/block/PhotoAtomicPrincipalCrossSectionBlock.python.cpp
whaeck: I've added this to the documentation.

src/ACEtk/PhotoAtomicTable.hpp
whaeck: I'm not sure but I would bet all of them are at a zero temperature.

src/ACEtk/PhotoAtomicTable/src/ctor.hpp
whaeck: You know me so well ...

src/ACEtk/PhotoAtomicTable/src/generateData.hpp
whaeck: The second PR on photoatomic data adds the Z value (as it stands in this PR, the current PhotoatomicTable is incomplete).

whaeck: Oh, and the user needs to provide the value for Z in the final version of this table type.

src/ACEtk/PhotoAtomicTable/src/generateData.hpp
whaeck: Not sure, we should talk more about this.

src/ACEtk/block/CoherentFormFactorBlock.hpp
whaeck: Yes, I need to get that done before the end of the FY.

PR #65, opened 2022-06-08
Feature/photoatomic 2

whaeck:
Finishing the photoatomic data table:
- implementing the optional data block
- additional testing


python/src/PhotoatomicTable.python.cpp
```diff
@@ -50,7 +59,10 @@ void wrapPhotoAtomicTable( python::module& module, python::module& ) {
     "    eszg      the principal cross section block\n"
     "    jinc      the incoherent scattering function block\n"
     "    jcoh      the coherent form factor block\n"
-    "    lhnm      the heating numbers block"
+    "    lhnm      the heating numbers block\n"
+    "    lhnm      the fluorescence data block\n"
```
cjosey, line 63: Comment is incorrect.

python/src/block/PhotoatomicComptonProfileBlock.python.cpp
```diff
@@ -0,0 +1,100 @@
...
+    "Initialise the block\n\n"
+    "Arguments:\n"
+    "    self        the block\n"
+    "    profiles    the compton profile data"
+  )
+  .def_property_readonly(
+
+    "NSH",
+    &Block::NSH,
+    "The number of electron shelld"
```
cjosey, line 48: Extraneous `d`.

python/src/block/PhotoatomicElectronShellBlock.python.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  // wrap views created by this block
+
+  // create the block
+  python::class_< Block > block(
+
+    module,
+    "PhotoatomicElectronShellBlock",
+    "The photoatomic LNEPS, LBEPS and LPIPS block with the electron shell data\n\n"
+    "The PhotoatomicPrincipalCrossSectionBlock class contains 3 arrays of the\n"
```
cjosey, line 28: Photoatomic Principal Cross Section Block is not correct here.

src/ACEtk/PhotoatomicTable.hpp
```diff
@@ -17,23 +20,26 @@ namespace ACEtk {
...
+ *  The PhotoatomicTable class represents the ACE table for the photoatomic data.
  */
-class PhotoAtomicTable : protected Table {
+class PhotoatomicTable : protected Table {
 
   /* fields */
   block::ESZG eszg_;
   block::JINC jinc_;
   block::JCOH jcoh_;
+  block::JFLO jflo_;
```
cjosey, line 31: Technically, JFLO, EPS, and SWD are optional, but I'm not worried much about that.

Edit: I see, they're just length zero if the NXS values say they're not there.

src/ACEtk/PhotoatomicTable/src/generateBlocks.hpp
```diff
@@ -17,7 +17,7 @@ auto block( std::size_t index ) const {
       end = *iter;
     }
   }
-
+std::cout << index << ' ' << start << ' ' << end << std::endl;
```
cjosey, line 20: I suspect this is a leftover debug message.

cjosey (approved):

python/src/PhotoatomicTable.python.cpp
whaeck: Yeah, that should be jflo

python/src/block/PhotoatomicComptonProfileBlock.python.cpp
whaeck: Fat fingers. That second d should have been an `s`

src/ACEtk/PhotoatomicTable.hpp
whaeck: I've probably replaced these with optionals later anyway.

Issue #67, opened 2022-07-07
Duplicate Writing of MT Number


abnowack:
I'm looking to zero out the inelastic collisions of U235. I've noticed that when changing MT=59 in the CrossSectionBlock, MT numbers will be altered in ways I don't believe it should.

[I've uploaded an example as a gist here](https://gist.github.com/abnowack/aef469c2b480b4764148405c62cedfcb)

You can see that 16 is removed, and 15 is added, in the reaction numbers, and additionally a duplicate 58 has been added.

In another script, when I go to read MT=59 in this resulting edited cross section file, I will get an error of "[error] The requested reaction number MT59 is not present" despite it appearing in the list of MT numbers.

This is using the ENDF data distributed with MCNP6.2, which I can attach or email privately if needed.

If I try to load this cross section in MCNP, I will get the following error,
```
 Expire parameter is 
 error 2 in  92235.80c of cross-section file xdata/endf71x/u/92235.710nc

 bad trouble in subroutine sread of xact                              

 error 2 in  92235.80c of cross-section file xdata/endf71x/u/92235.710nc
```

whaeck:
Nice catch. It's actually a coding "flaw", not an error in the ACE file. Allow me to explain.

The ACE file is being printed out as a list of floating point values. MCNP actually reads these into an array of floating point values, and interprets these numbers as integers as required. As you probably know, converting a floating point value into an integer value by implicit casting can be dangerous since some integers cannot be represented properly in a floating point representation (implicit casts tend to round down to the closest integer). In this case, the ACE file contains the value 5.90000000000e+1, which gets cast to 58 because the in memory representation is probably 5.899999999999e+1 (or something similar). In order to extract the correct integer, we actually need to round the float to the closest integer before casting the float to an integer value.

In this case, the C++ function under the hood returns an array of floating points, which on the python side get cast to an array of integer values - which is bad. I actually have to do the rounding on the C++ side.

You can see this behaviour in the following snippet of code:
```
for i in range( 1, ace_file2.NTR + 1 ) :
    
    print( ace_file2.MTR.reaction_number(i), ace_file2.MTR.reaction_numbers[i-1] )
```

The `ace_file2.MTR.reaction_number(i)` does the rounding while `ace_file2.MTR.reaction_numbers[i-1]` only casts under the hood. As you can see when you execute this piece of python code, even the first MT number gets interpreted wrongly: the correct value is 16, not 15.

I need to fix this asap.

[...]

PR #68, opened 2022-07-07
Fix/rounding

whaeck:
Fixing some more rounding issues, following issue #67


src/ACEtk/block/SecondaryParticleTypeBlock.hpp
```diff
@@ -82,9 +82,14 @@ class SecondaryParticleTypeBlock : protected details::ArrayData {
...
 
-    return std::find( this->begin(),
-                      this->begin() + this->NTYPE(),
-                      type ) != this->end();
+    return std::find_if(
+               this->begin(),
+               this->begin() + this->NTYPE(),
+               [type] ( auto&& value ) {
+
+                 return static_cast< unsigned int >( std::round( value ) )
```
cjosey, line 90: As an aside, I have found in our code that rounding to nearest (`std::round`) instead of rounding to zero `static_cast<int>` has a substantial performance penalty. However, our code has taken the stance that input is valid and was checked ahead of time for performance reasons, whereas this code cannot.

cjosey (dismissed):

src/ACEtk/block/SecondaryParticleTypeBlock.hpp
whaeck: ACEtk should not be used during simulations so performance is less of an issue - although it is a valid point I should look into.

cjosey (approved):

PR #70, opened 2022-07-28
Feature/optional

whaeck:
Using std::optional to represent optional blocks in ACE files instead of relying on empty()

### DIFF
--- a/src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
+++ b/src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
@@ -452,12 +452,12 @@ void verifyChunkU235( const ContinuousEnergyTable& chunk ) {
   CHECK( 76518 == chunk.ESZ().XSS( 1, chunk.NES() ).size() );
 
   // NU block
-  CHECK( false == chunk.NU().empty() );
+  CHECK( false == chunk.NU()->empty() );
 
-  CHECK( true == chunk.NU().hasPromptAndTotalFissionMultiplicity() );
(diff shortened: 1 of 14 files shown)


src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
```diff
@@ -688,34 +688,34 @@ void verifyChunkU235( const ContinuousEnergyTable& chunk ) {
...
+  CHECK( true == std::holds_alternative< OutgoingEnergyDistributionData >( chunk.DNED()->energyDistributionData( 6 ) ) );
 
-  CHECK( std::nullopt == chunk.DNED().referenceFrame( 1 ) );
-  CHECK( std::nullopt == chunk.DNED().referenceFrame( 6 ) );
+  CHECK( std::nullopt == chunk.DNED()->referenceFrame( 1 ) );
+  CHECK( std::nullopt == chunk.DNED()->referenceFrame( 6 ) );
 
   // PTYPE block
-  CHECK( true == chunk.PTYPE().empty() );
+  CHECK( std::nullopt == chunk.PTYPE() );
```
cjosey, line 712: I think you may have done this in the future-past, but this would be a good use of `has_value()`.

cjosey (dismissed):

src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
whaeck: Yes, I have changed stuff like this - although I might have forgotten to look at tests

cjosey (approved):

PR #71, opened 2022-07-29
Feature/dn number

whaeck:
Delayed neutron precursor groups now have their group number stored as well.

### DIFF
--- a/python/src/block/DelayedNeutronPrecursorData.python.cpp
+++ b/python/src/block/DelayedNeutronPrecursorData.python.cpp
@@ -33,32 +33,41 @@ void wrapDelayedNeutronPrecursorData( python::module& module, python::module& ) 
   block
   .def(
 
-    python::init< double, std::vector< double >, std::vector< double > >(),
-    python::arg( "constant" ), python::arg( "energies" ),
-    python::arg( "probabilities" ),
+    python::init< unsigned int, double,
+                  std::vector< double >, std::vector< double > >(),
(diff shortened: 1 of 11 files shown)


python/src/block/DelayedNeutronPrecursorData.python.cpp
```diff
@@ -33,33 +33,42 @@ void wrapDelayedNeutronPrecursorData( python::module& module, python::module& )
...
-    python::arg( "constant" ), python::arg( "energies" ),
-    python::arg( "probabilities" ),
+    python::init< unsigned int, double,
+                  std::vector< double >, std::vector< double > >(),
+    python::arg( "number" ), python::arg( "constant" ),
+    python::arg( "energies" ), python::arg( "probabilities" ),
     "Initialise the block assuming linear interpolation\n\n"
     "Arguments:\n"
     "    self             the block\n"
+    "    number           the precursor group number\n"
```
cjosey, line 43: Is it expected that groups will be sparse or stored out-of-order?

cjosey (dismissed):
I'm not wholly confident as to what marginal benefit this has, but nothing seems amiss.

python/src/block/DelayedNeutronPrecursorData.python.cpp
whaeck: I was planning on sorting the groups in order but for some reason I decided not to. I forgot why.

cjosey (approved):

PR #72, opened 2022-08-11
Feature/header

whaeck:
The NJOY header and 2.0.x headers are now fully supported


python/src/Header201.python.cpp
```diff
@@ -0,0 +1,101 @@
...
+    python::init< const std::string&, const std::string&,
+                  double, double, const std::string&,
+                  const std::vector< std::string >& >(),
+    python::arg( "szaid" ), python::arg( "source" ), python::arg( "awr" ),
+    python::arg( "temperature" ), python::arg( "date" ),
+    python::arg( "comments" ),
+    "Initialise the component\n\n"
+    "Arguments:\n"
+    "    self          the header\n"
+    "    szaid         the szaid (max 10 characters)\n"
```
cjosey, line 40: I think the szaid in the new format is allowed to be 24 characters.

src/ACEtk/Table/Header201.hpp
```diff
@@ -0,0 +1,104 @@
...
+ *              SZAID (24 characters), blank space,
+ *              SRC (evaluation source, 24 characters)
+ *    - line 2: AW (atomic mass ratio, 12 characters), blank space,
+ *              TZ (temperature in MeV, 12 characters), blank space,
+ *              date (10 characters), blank space, N (number of comment lines,
+ *              4 characters)
+ *    - line 3 to N+2: comment lines (the last two lines are always the normal
+ *                     header)
+ */
+class Header201 {
```
cjosey, line 29: To be honest, I am unclear why there are separate Header and Header201 classes. The data they contain is the same (201 being a superset) with perhaps the sole exception of the ZAID. Further, Lib80x files have both headers.

cjosey (commented):
I think that I'm going to pause review of this one for further discussion in person and go on to the next few for now.

python/src/Header201.python.cpp
whaeck: Yes.

src/ACEtk/Table/Header201.hpp
whaeck: By convention, they do but the last two lines being the old header was never "standardised" so they do contain different information (I think the title is no present in the 201 header).

cjosey (dismissed):
We have discussed these changes, and improvements will be made in a future PR once the history is consistent.

cjosey (approved):

PR #73, opened 2022-09-21
Feature/table ctor

whaeck:
Enabling the table ctor using a header and a data object.

Added Python test for Header and Header201, updated Table test.


cjosey (approved):
Same comment as previous PR.

PR #80, opened 2023-03-13
Changing following review #46

whaeck:
These contain all changes following review of PR #46

### DIFF
--- a/src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
+++ b/src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
@@ -8,7 +8,7 @@ generateXSS( const std::string& name,
   // add the distribution data to the xss array
   const auto nr = boundaries.size();
   const auto ne = distributions.size();
-  std::vector< double > xss( static_cast< double >( 1 + 2 * ne ) );
+  std::vector< double > xss( 1 + 2 * ne );
   xss[0] = ne;
   std::size_t index = 1;
   const auto offset = 1 + 2 * nr + locb;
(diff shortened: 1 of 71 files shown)


whaeck:
@cjosey I think we're ready on this one.

src/ACEtk/block/details/BaseDistributionData/src/generateXSS.hpp
```diff
@@ -6,19 +6,19 @@ generateXSS( const std::string& name,
              std::size_t locb ) {
 
   // add the distribution data to the xss array
-  std::size_t nr = boundaries.size();
-  std::size_t ne = distributions.size();
-  std::vector< double > xss( 1 + 2 * ne );
+  const auto nr = boundaries.size();
+  const auto ne = distributions.size();
+  std::vector< double > xss( static_cast< double >( 1 + 2 * ne ) );
```
cjosey: This appears in a few places, `xss` is constructed using a double instead of what previously was an int/size_t, which is passed to the size. I think that's the only thing I've seen in this PR that is an issue.

cjosey: Approved once these are all fixed.

whaeck: I did not pay attention here it would appear. This should only be done in the initialiser list, e.g. 
```
xss = { static_cast<double>(...) }
```
I can fix this.

whaeck: I found 7 places where this had happened. These are fixed now.

cjosey (dismissed):

cjosey (approved):

PR #92, opened 2023-06-02
Updating pybind11 dependecy version to resolve a compilation issue

whaeck:
@cjosey This should resolve that compilation issue you mentioned.

### DIFF
--- a/cmake/develop_dependencies.cmake
+++ b/cmake/develop_dependencies.cmake
@@ -54,7 +54,7 @@ FetchContent_Declare( range-v3
 
 FetchContent_Declare( pybind11
     GIT_REPOSITORY  https://github.com/pybind/pybind11
-    GIT_TAG         v2.10.1
+    GIT_TAG         v2.10.4
     )
 
 #######################################################################
--- a/cmake/release_dependencies.cmake
+++ b/cmake/release_dependencies.cmake
@@ -32,7 +32,7 @@ FetchContent_Declare( header-utilities
 
 FetchContent_Declare( interpolation
     GIT_REPOSITORY  https://github.com/njoy/interpolation
-    GIT_TAG         32375a9ae0ff346cecf69d5acd32dbc0092e7d7a
+    GIT_TAG         cb975c3cc58ccbc948b6750231056f7a98ad2ac9
     )
 
 FetchContent_Declare( Log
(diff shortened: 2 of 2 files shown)


cjosey:
It does not resolve GCC 13 (I think because one of the dependencies added `-Werror` -- this is why `-Werror` should never be in production because warnings change with time), but it does resolve LLVM 16.

cjosey (dismissed):

whaeck:
I didn't know there was a gcc 13 issue. What exactly is it?

cjosey:
```
In file included from [...]/ACEtk/buildgcc/_deps/range-v3-src/include/range/v3/range_fwd.hpp:20,
                 from [...]/ACEtk/buildgcc/_deps/range-v3-src/include/range/v3/view/any_view.hpp:22,
                 from [...]/ACEtk/python/src/views.hpp:10,
                 from [...]/ACEtk/python/src/ACEtk.python.cpp:6:
[...]/ACEtk/buildgcc/_deps/range-v3-src/include/meta/meta.hpp:3190:19: error: declaration of 'template<class Fn> template<class State, class A> using meta::detail::partition_<Fn>::invoke = meta::_t<meta::detail::partition_<Fn>::impl<State, A> >' changes meaning of 'invoke' [-Wchanges-meaning]
 3190 |             using invoke = _t<impl<State, A>>;
      |                   ^~~~~~
[...]/ACEtk/buildgcc/_deps/range-v3-src/include/meta/meta.hpp:3185:34: note: used here to mean 'using meta::invoke = typename Fn::invoke<A>'
 3185 |                 using type = if_<invoke<Fn, A>, pair<list<Yes..., A>, list<No...>>,
      |                                  ^~~~~~~~~~~~~
[...]/ACEtk/buildgcc/_deps/range-v3-src/include/meta/meta.hpp:541:11: note: declared here
  541 |     using invoke = typename Fn::template invoke<Args...>;
      |           ^~~~~~
make[2]: *** [CMakeFiles/ACEtk.python.dir/python/src/ACEtk.python.cpp.o] Error 1
make[1]: *** [CMakeFiles/ACEtk.python.dir/all] Error 2
make: *** [all] Error 2
```

whaeck:
Ugh. I hate it when that happens. We were planning on ripping out ranges to enable intel support, so ...

cjosey (approved):

PR #99, opened 2023-08-30
Feature/eprdata

whaeck:
Adding two blocks for the eprdata files: the subshell block (from the atomic relaxation sublibrary) and the electron cross section data block (with the electron cross section data from the electron sublibrary)


cjosey (dismissed):

PR #100, opened 2023-08-31
Feature/eprdata part2

whaeck:
This adds the EXCIT and BREML blocks for the eprdata format. Both are tabulated data as a function of energy (excitation energy loss for EXCIT and average remaining energy for the electron after Bremsstrahlung for BREML).

I also updated the eprdata test files. The previous files had incorrect formatting because the atom editor tries to be too intelligent.


cjosey (dismissed):

PR #101, opened 2023-09-06
Feature/eprdata part3

whaeck:
Extending the JINC and JCOH blocks to use arbitrary tabulated data (introduced in MCNPX and the eprdata files for MCNP).

This also adds the SUBSH block to the PhotoatomicTable and adds the first eprdata12 test for it.


cjosey (dismissed):

PR #102, opened 2023-09-11
Feature/eprdata part4

whaeck:
Adding the photoelectric cross section data block, giving access to the electron cross section, electron excitation energy loss and Bremsstrahlung blocks.

Further testing using the content of the eprdata12 file.


cjosey (dismissed):

PR #103, opened 2023-09-18
Feature/eprdata part5

whaeck:
This adds the XPROB block with the subshell transition data.

Fun fact: this is the ONLY block in the ACE format that I know off that does not use locators but offsets. Sigh. As a result, I updated the BaseBlockWithLocators class to work both for offsets and locators. As a result, the XPROB interface abstracts the offsets away by using a locator based interface. In the XSS array, we do still have offsets for the XPROB block.

edit: so the elastic angular distribution tables and the bremsstrahlung photon energy distributions do use offsets too. Cue more sighing.


src/ACEtk/block/PhotoatomicSubshellTransitionData/src/ctor.hpp
```diff
@@ -0,0 +1,32 @@
+PhotoatomicSubshellTransitionData() = default;
+
+PhotoatomicSubshellTransitionData( const PhotoatomicSubshellTransitionData& ) = default;
+PhotoatomicSubshellTransitionData( PhotoatomicSubshellTransitionData&& ) = default;
+
+/**
+ *  @brief Constructor
+ *
+ *  @param[in] energies    the energy values
+ *  @param[in] loss        the average excitation energy loss values
```
cjosey, line 10: This constructor documentation is inconsistent with its contents. This might be fixed in a future PR (but really, this is why you don't do 10 PRs on one topic).

whaeck: That has been fixed in #112. By then this was changed to src/ACEtk/electron/SubshellTransitionData/src/ctor.hpp

whaeck: We all know why we try to do it this way ...

cjosey (approved):

PR #104, opened 2023-09-20
Feature/eprdata part6

whaeck:
This adds the ELAS electron elastic angular distribution block.

An additional base class is added for the recurring eprdata block with energy values, lenghts, off sets and tables.

This base class abstracts away the use of these offsets and replaces them with locators to get a uniform feel over the entire ACE format.


cjosey (dismissed):

PR #105, opened 2023-09-20
Feature/eprdata part7

whaeck:
Adding the BREME block for the outgoing photon energy distributions from Bremsstrahlung

Note: the next PR will rename the ElectronBremstrahlungEnergyDistributionBlock to ElectronEnergyDistributionBlock since the knock-on electron energy distributions for each subshell use the same structure.

Also: the "Electron" in the front of this name is deceptive: the block applies to both secondary photons for the BREME block and the secondary knock-on electrons in the EION blocks. This will be resolved later when we disentangle the block namespace into several other namespaces. THe block namespace is getting out of hand.


cjosey (dismissed):

PR #107, opened 2023-09-25
Feature/eprdata part8

whaeck:
Adding the EION block and eprdata14 testing

This includes all eprdata12 features and we can now create eprdata12 formatted files.


src/ACEtk/PhotoatomicTable/test/resources/1000.12p
```diff
@@ -6,7 +6,7 @@ ENDF/B-VI Release 8 photo/electro-atomic and relaxation data for 1-H     mat 100
       0         0.      0         0.      0         0.      0         0.
     12025        1      647        0        1        1        1      342
       170       16       10       82        0        0        0        0
-        1     3236     3424     3694     3694     4341     4342     4343
+        1     3236     3424        0     3694     4341     4342     4343
```
cjosey, line 9: As discussed via email, this is likely going to cause a problem.

whaeck: yes. I'll have to put that back. It's currently in #115

cjosey (approved):

PR #108, opened 2023-09-26
Feature/eprdata part9

whaeck:
The update for eprdata14:
- the addition of the SELAS block with elastic scattering cross section information
- the NXS(13) and NXS(14) values - not that we needed these

And thus ends the quest to understand and support the eprdata format!


cjosey (dismissed):

PR #110, opened 2023-09-27
Feature/refactor block

whaeck:
This is part 1 of a major ACEtk refactor. This refactor consists in splitting up the existing block namespace into separate namespaces by data type:
- continuous: for incident neutron and charged particles
- photonuclear
- photoatomic: photon related blocks only (electron data in eprdata gets its own namespace)
- electron
- thermal
- dosimetry

Eventually, we can add a multigroup one too.

This PR introduces the "photoatomic" namespace. The build system was slightly updated.


cmake/unit_testing_python.cmake
```diff
@@ -0,0 +1,32 @@
...
+
+function( add_python_test name source )
+
+  set( test_name "ACEtk.python.${name}.test" )
+  add_test( NAME ${test_name}
+            COMMAND ${PYTHON_EXECUTABLE} -m unittest -v test/${source}
+            WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/python )
+  set_tests_properties( ${test_name}
+                        PROPERTIES ENVIRONMENT
+                        PYTHONPATH=${ACEtk_PYTHONPATH}:$ENV{PYTHONPATH})
```
cjosey, line 15: I can confirm this approach works for all tests linked into this function. If you want to change your CMake minimum version to 3.22, you can change this to `PROPERTIES ENVIRONMENT_MODIFICATION PYTHONPATH=string_prepend:${ACEtk_PYTHONPATH}`. But that's a silly thing to update your minimum CMake Version over.

I hope there is intention to port this to the rest of the tests and to do something similar with the C++ ones.

cjosey (dismissed):
This is a pretty rote refactor. One suggestion for future work, otherwise approval.

cmake/unit_testing_python.cmake
whaeck: I indeed want to do the same for the C++ tests but that involves moving to Catch2 version 3. If you want to see what it looks like, look here: https://github.com/njoy/scion/blob/feature/integration-part5/cmake/unit_testing.cmake

whaeck: Once I finsih the namespace refactor I'll move to the C++ tests.

PR #111, opened 2023-09-27
Feature/refactor block part2

whaeck:
All electron related data is now in the electron namespace.


python/src/PhotoatomicTable.python.cpp
```diff
@@ -21,18 +21,18 @@ void wrapPhotoatomicTable( python::module& module, python::module& ) {
   using JCOH = njoy::ACEtk::photoatomic::JCOH;
   using JFLO = njoy::ACEtk::photoatomic::JFLO;
   using LHNM = njoy::ACEtk::photoatomic::LHNM;
-  using EPS = njoy::ACEtk::block::EPS;
+  using EPS = njoy::ACEtk::electron::EPS;
```
cjosey, line 24: General question on design. This is an electron block in the electron-photon-relaxation data. If we add ability to construct new condensed history data, would this pose a conflict?

cjosey (approved):
Another simple refactoring.

python/src/PhotoatomicTable.python.cpp
whaeck: I would not think so. The distinction goes by incident particle type essentially.

PR #114, opened 2023-09-28
Feature/refactor block part4

whaeck:
This moves the dosimetry blocks to the dosimetry namespace.

Up next: thermal scattering


cjosey (approved):
Another simple refactor.

whaeck:
But it'll be sooooo much better once all is done ;-)

PR #117, opened 2023-09-30
Feature/refactor block part7

whaeck:
this introduces the continuous namespace.


src/ACEtk/continuous/FissionMultiplicityBlock/test/CMakeLists.txt
cjosey, line 1: I just wanted to comment on how amusing it is that Git just fails to identify where this was moved from because the edit confused it. This seems to have happened with almost all the CMakeLists.txt files. It's not important as, since the code built and the tests run I know they're all in the right place.

cjosey (dismissed):
Well, that was a fun review. Clicking "Load Diff" 250 times was a pain. But everything looks good, no surprises.

src/ACEtk/continuous/FissionMultiplicityBlock/test/CMakeLists.txt
whaeck: I know. I even try to work around that by renaming/moving the files first, and then committing just the name change in the hopes of it getting picked up as being the same file. Sometimes it works, sometimes it doesn't.

I merged the previous PR so this is now stale but I'm going to ignore that and just admin merge.

PR #118, opened 2023-10-02
Feature/catch2 upgrade

whaeck:
Updating ACEtk unit tests to use Catch2 v3, removing the outdated catch-adapter dependency

### DIFF
--- a/cmake/develop_dependencies.cmake
+++ b/cmake/develop_dependencies.cmake
@@ -11,9 +11,9 @@ FetchContent_Declare( Log
     GIT_SHALLOW     TRUE
     )
 
-FetchContent_Declare( catch-adapter
-    GIT_REPOSITORY  https://github.com/njoy/catch-adapter
-    GIT_TAG         origin/master
+FetchContent_Declare( Catch2
+    GIT_REPOSITORY  https://github.com/catchorg/Catch2
+    GIT_TAG         v3.3.2
     GIT_SHALLOW     TRUE
     )
 
@@ -39,7 +39,7 @@ FetchContent_Declare( pybind11
 
 FetchContent_MakeAvailable(
     Log
-    catch-adapter
+    Catch2
     disco
     range-v3
     pybind11
--- a/src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
+++ b/src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
@@ -1,10 +1,13 @@
-#define CATCH_CONFIG_MAIN
+// include Catch2
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+using Catch::Matchers::WithinRel;
 
-#include "catch.hpp"
-#include "ACEtk/fromFile.hpp"
+// what we are testing
 #include "ACEtk/ContinuousEnergyTable.hpp"
 
 // other includes
+#include "ACEtk/fromFile.hpp"
 
 // convenience typedefs
 using namespace njoy::ACEtk;
@@ -50,7 +53,7 @@ SCENARIO( "ContinuousEnergyTable" ){
         CHECK( iz.size() == iz_chunk.size() );
         for ( unsigned int i = 0; i < iz_chunk.size(); ++i ) {
 
-          CHECK( iz[i] == Approx( iz_chunk[i] ) );
+          CHECK( iz[i] == iz_chunk[i] );
         }
       } // THEN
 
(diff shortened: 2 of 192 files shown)


cmake/develop_dependencies.cmake
```diff
@@ -11,9 +11,9 @@ FetchContent_Declare( Log
     GIT_SHALLOW     TRUE
     )
 
-FetchContent_Declare( catch-adapter
-    GIT_REPOSITORY  https://github.com/njoy/catch-adapter
-    GIT_TAG         origin/master
+FetchContent_Declare( Catch2
+    GIT_REPOSITORY  https://github.com/catchorg/Catch2
+    GIT_TAG         v3.3.2
```
cjosey, line 16: Why 3.3.2 instead of 3.4.0?

src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
```diff
@@ -386,7 +389,7 @@ SCENARIO( "ContinuousEnergyTable" ){
         CHECK( jxs.size() == jxs_chunk.size() );
         for ( unsigned int i = 0; i < jxs_chunk.size(); ++i ) {
 
-          CHECK( jxs[i] == Approx( jxs_chunk[i] ) );
+          CHECK( jxs[i] == jxs_chunk[i] );
```
cjosey, line 392: [suggestion, untested]

It appears that Catch2 v3 has a number of range-equals (or All) capabilities that optionally allow you to set the comparator. So it may be possible to construct a check if vectors approximately equal one another without the loop.

src/ACEtk/ThermalScatteringTable/test/ThermalScatteringTable.test.cpp
```diff
@@ -629,7 +632,7 @@ void verifyChunkH2O( const ThermalScatteringTable& chunk ) {
 void verifyChunkZrZrH( const ThermalScatteringTable& chunk ) {
 
   CHECK( "zr-zrh.40t" == chunk.ZAID() );
-  CHECK( 2.53e-8 == Approx( chunk.temperature() ) );
+  CHECK_THAT( 2.5507e-8, WithinRel( chunk.temperature() ) );
```
cjosey, line 635: Any backstory as to why this isn't room temp?

src/ACEtk/ThermalScatteringTable/test/ThermalScatteringTable.test.cpp
```diff
@@ -780,7 +783,7 @@ void verifyChunkAl( const ThermalScatteringTable& chunk ) {
 void verifyChunkDLiD( const ThermalScatteringTable& chunk ) {
 
   CHECK( "d-lid.10t" == chunk.ZAID() );
-  CHECK( 2.53e-8 == Approx( chunk.temperature() ) );
+  CHECK_THAT( 3.4469e-8, WithinRel( chunk.temperature() ) );
```
cjosey, line 786: Hah, guess `Approx` was too approx.

src/ACEtk/continuous/AngleEnergyDistributionData/test/AngleEnergyDistributionData.test.cpp
```diff
@@ -166,54 +169,54 @@ void verifyChunk( const AngleEnergyDistributionData& chunk ) {
...
   CHECK( 2 == data12.pdf().size() );
-  CHECK( .5 == Approx( data12.pdf().front() ) );
-  CHECK( .5 == Approx( data12.pdf().back() ) );
+  CHECK_THAT( .5, WithinRel( data12.pdf().front() ) );
+  CHECK_THAT( .5, WithinRel( data12.pdf().back() ) );
 
   CHECK( 2 == data12.cdf().size() );
-  CHECK( 0. == Approx( data12.cdf().front() ) );
-  CHECK( 1. == Approx( data12.cdf().back() ) );
+  CHECK_THAT( 0., WithinRel( data12.cdf().front() ) );
```
cjosey, line 202: [aside]

Is WithinRel useful for zero? There is no non-zero element on the end where a match will occur. I don't think it's worth modifying this, though, as 0 tends to have "meaning" and would be bad to be wrong on.

cjosey (approved):
Took longer to review due to Github 404ing frequently, but good to go.

cmake/develop_dependencies.cmake
whaeck: For the moment, I just want all njoy repos to use the same dependencies. And scion is currently on this version - when I updated scion 3.3.2 was the latest version.

Once all njoy repos are up to snuff, we can start updating dependencies consistently.

src/ACEtk/ContinuousEnergyTable/test/ContinuousEnergyTable.test.cpp
whaeck: True, but does it tell you which element is wrong if it goes wrong?

src/ACEtk/ThermalScatteringTable/test/ThermalScatteringTable.test.cpp
whaeck: because for some people 296 K is RT?

src/ACEtk/ThermalScatteringTable/test/ThermalScatteringTable.test.cpp
whaeck: And this is why this update was useful!

PR #120, opened 2023-10-11
Feature/update cmake

whaeck:
Some updates to the CMakeLists.txt file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ project( ACEtk
   LANGUAGES CXX
 )
 
+include( CTest )
 include( CMakeDependentOption )
 include( GNUInstallDirs )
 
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -11,18 +25,16 @@ project( ACEtk LANGUAGES CXX )
...
 
-option( ACEtk.tests "Build the ACEtk unit tests" OFF )
-option( ACEtk.python "Build ACEtk python bindings" ON )
-option( strict_compile
-    "Treat all warnings as errors." ON
-    )
-
-# Compile flags
-set( common_flags "-Wall" "-Wextra" "-Wpedantic" )
-set( strict_flags "-Werror" )
```
cjosey, line 22: As a note, these two still appear as `${PREFIX}_X_flags` in `src/legacy/interpretation`. I think they were already empty strings so it's no big deal, and `CMAKE_RELEASE_FLAGS` and similar do most of what is needed.

cjosey (approved):
All looks good.

CMakeLists.txt
whaeck: Once we update to use shacl cmake, we'll add the flags back in. For now, I just want to make sre that all njoy libraries use the same thing everywhere.

PR #122, opened 2024-02-06
Feature/consistent interface

whaeck:
A few small modifications to the interface of the photonuclear (PN), continuous energy (CE) and photoatomic (PA) tables:
- adding S(), Z(), A() where they were not available
- the S value is added to the constructors of the CE and PN tables
- some documentation updates

For the values of S, Z and A: if the CE and PN table do not have values in NXS(9), NXS(10) and NXS(11), then ACEtk will determine these values and them to the NXS array.


python/src/PhotonuclearTable.python.cpp
```diff
@@ -161,6 +162,42 @@ void wrapPhotonuclearTable( python::module& module, python::module& ) {
     &Table::numberSecondaryParticleTypes,
     "The number of secondary particle types"
   )
+  .def_property_readonly(
+
+    "S",
+    &Table::S,
+    "The excited state of the target"
```
cjosey: [Question]
Is there any need to disambiguate "excited" here? It could mean isomeric state or index of energy level.

src/ACEtk/ContinuousEnergyTable/src/generateData.hpp
```diff
@@ -330,6 +330,7 @@ Data generateData( unsigned int z, unsigned int a,
   nxs[5] = ntrp;
   nxs[6] = ntype;
   nxs[7] = npcr;
+  nxs[8] = s;
```
cjosey, line 336: Probably should add to comments unless you think this is too obvious. "The S is the S".

src/ACEtk/ContinuousEnergyTable/src/setSZA.hpp
```diff
@@ -0,0 +1,50 @@
...
+        a -= 400;
+        s = 1;
+        while ( a > estimate + 40 ) {
+
+          a -= 100;
+          ++s;
+        }
+      }
+
+      if ( ( z == 95 ) && ( a == 242 ) ) {
```
cjosey, line 40: Suggest documenting this as it is obscure (and then closing the associated issue :)

src/ACEtk/ContinuousEnergyTable/src/setSZA.hpp
```diff
@@ -0,0 +1,50 @@
...
+      unsigned int estimate =
+      static_cast< unsigned int >( std::round((0.0064*z+2.0009)*z-0.281));
+      if ( a > estimate + 40 ) {
+
+        a -= 400;
+        s = 1;
+        while ( a > estimate + 40 ) {
+
+          a -= 100;
+          ++s;
```
cjosey, line 34: This code is not safe for an input of, for example, 92279. On the first pass, A > estimate + 40. So it subtracts 400. But A is unsigned. And so now it is 4 billion, which results in this while loop getting stuck.

cjosey (changes requested):
Minor fixes I think are required. Rest looks good

src/ACEtk/ContinuousEnergyTable/src/setSZA.hpp
```diff
@@ -0,0 +1,50 @@
...
+    unsigned int z = ( za - a ) / 1000;
+    unsigned int s = 0;
+    if ( z > 999 ) {
+
+      unsigned int sz = z;
+      z = z % 1000;
+      s = ( sz - z ) / 1000;
+    }
+
+    if ( a != 999 ) {
```
cjosey: Also, sorry I misspoke when I was explaining what was in our internals. A Z of 999 is "nonphysical". We have test data of ZAID = 999902.01c and stuff.

python/src/PhotonuclearTable.python.cpp
whaeck: True, I should probably use metastable here.

src/ACEtk/ContinuousEnergyTable/src/generateData.hpp
whaeck: I sometimes have trouble seeing when you're joking or not :-)

src/ACEtk/ContinuousEnergyTable/src/setSZA.hpp
whaeck: And this is why you do not store two (or three) pieces of information in a single field: disentangling it is just a mess :-)

Now that I think of it, the original I had there (a > 3*z) would have had the same issue.

src/ACEtk/ContinuousEnergyTable/src/setSZA.hpp
whaeck: OK, so in this case, does it make sense to try and derive a proper A value? I can just change this line to if (z != 999) and be done with it.

src/ACEtk/ContinuousEnergyTable/src/generateData.hpp
cjosey: Well, it was just a matter of consistency. All the other values were documented, but I was admitting that this code barely needed documenting.

src/ACEtk/ContinuousEnergyTable/src/setSZA.hpp
cjosey: Yes this is good.

cjosey (approved):

PR #125, opened 2024-03-21
Fix/xsdir comment lines

whaeck:
Added logic to read comment lines in xsdir files.


src/ACEtk/XsdirEntry/src/parse.hpp
```diff
@@ -1,23 +1,54 @@
+static void trim( std::string& string ) {
+
+  // useful lambdas
+  auto isNotSpace = [] ( const auto& c ) { return !std::isspace( c ); };
```
cjosey, line 4: I'm just going to leave this comment now so I don't forget in the morning.

There's two versions of `std::isspace`, and I'm not sure which you're getting since `XsdirEntry.hpp` doesn't contain either. If it came from `<cctype>` then this probably needs to be:
```c++
auto isNotSpace = [] ( const unsigned char c ) { return !std::isspace( c ); };
```
As `isspace` is only well-defined for `unsigned char` and `string` may contain `char` (at least, [so `cppreference` claims](https://en.cppreference.com/w/cpp/string/byte/isgraph)).

If it's coming from `<locale>` then no changes are required, though I probably would replace it with `std::isgraph()`.

src/ACEtk/Xsdir/src/parse.hpp
```diff
@@ -22,46 +22,70 @@ static void trimToLowerCase( std::string& string ) {
...
 }
 
+static std::string getline( std::istream& in, std::istream::pos_type& position ) {
+
+  // a line is a comment line if the first character on it is '#'
+  auto isCommentLine = [] ( const auto& line ) {
+
+    if ( line.size() > 0 ) {
+
+      return line[0] == '#' ? true : false;
```
cjosey: This line is identical to `return line[0] == '#'`. I don't imagine this is more particularly readable. Though, I admit I'd rewrite this as:
```cpp
auto isCommentLine = [] (const auto& line) {
    return (line.size() > 0 && line[0] == '#');
}
```

src/ACEtk/Xsdir/src/parse.hpp
```diff
@@ -22,46 +22,70 @@ static void trimToLowerCase( std::string& string ) {
...
+      current = getline( in, position );
+      toLowerCase( current );
     }
   }
 
-  // read a line and verify for the atomic weight ratios
-  std::getline( in, current );
-  trimToLowerCase( current );
+  // verify for the atomic weight ratios
   if ( current != "atomic weight ratios" ) {
```
cjosey, line 89: I don't think this is worth adding support for, but the file technically can have these blocks in any order. It may be worth tightening the MCNP definition of the file.

src/ACEtk/Xsdir/test/Xsdir.test.cpp
```diff
@@ -196,6 +221,30 @@ std::string chunk() {
...
+    "    1000   0.99931697    1001   0.99916733    1002   1.99679968    1003   2.99013997\n"
+    "                         1004   3.99320563\n"
+    "# this is a comment line\n"
+    "   92000 235.98412800   92235 233.02478975 1092235 233.02478975\n"
+    "\n"
+    "# this is a comment line\n"
+    "directory\n"
+    "# this is a comment line\n"
+    " 92234.00c   234.000000 file 0 1 5 1000 0 0 2.530000e-08\n"
+    " 92235.00c   235.000000 file2 0 1 3 2000 0 0 2.530000e-08 ptable\n"
```
cjosey: There is no test of a continue line.

cjosey (dismissed):
Not much to remark on. There are a few missing pieces in the test. It may also be worthwhile to see if your code safely ignores the `obsolete` token.

src/ACEtk/Xsdir/src/parse.hpp
whaeck: I just like my ternary statements :-)

I might actually change this since it can be shortened as you point out.

src/ACEtk/Xsdir/src/parse.hpp
whaeck: I did not know this. That explains why you use a state machine like code in MCNP.

src/ACEtk/Xsdir/test/Xsdir.test.cpp
whaeck: Good point. I will add one.

whaeck: There actually is one in the XsdirEntry test - but it never hurts to add one to Xsdir too.

src/ACEtk/XsdirEntry/src/parse.hpp
whaeck: Interesting. I did not know this. I assume it is <locale> since I use standard C++ library files. I added the <locale> include just in case.

whaeck:
@cjosey Ready for re-review following changes.

cjosey (approved):
Changes all look good.

PR #126, opened 2024-03-21
A few changes to cmake

whaeck:


cjosey (dismissed):
Simple change, all looks good.

PR #127, opened 2024-03-26
Fix/msvc

whaeck:
A few fixes to ACEtk so that it compiles using MSVC.

### DIFF
--- a/src/ACEtk/Table/Data.hpp
+++ b/src/ACEtk/Table/Data.hpp
@@ -40,52 +40,52 @@ public:
   /**
    *  @brief Return the IZ array
    */
-  const auto& IZ() const { return this->izaw_.first; }
+  const std::array< int32_t, 16 >& IZ() const { return this->izaw_.first; }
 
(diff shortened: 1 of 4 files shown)


cjosey:
I'll review this shortly, but did the `/permissive-` flag help any?

Edit: Actually, it seems that that flag is now the default in MSVC2019 onward. I wonder if that includes CMake targets.

cjosey (approved):
I will note that I cannot test this with MSVC, as I don't have access to a machine that has the compilers right now, but I can guess as to the errors. MSVC can't handle `auto` return from a `decltype(auto)` return and its STL headers are better about not including each other? Was this reported from 2019 or 2022? Reading the Range-V3 issue tracker, there may be a 2022 issue with that too.

I can confirm at least there's no issues building with LLVM 16 or IntelLLVM 2024.0.2 (albeit with lots and lots of warnings due to spdlog). The issue with building with GCC 13 is still present.

PR #128, opened 2024-04-23
Update/tools dependency

whaeck:
This removes range-v3 and disco as a dependency from ACEtk and replaces it with components from the tools library (NanoRange for range-v3 and a rewrite of disco for disco). It also removed binding the common AnyView types (for int, double, etc.) by moving these to tools as well (this removes the issue of binding the types twice in ENDFtk and ACEtk).

The interface has undergone minimal changes so the impact of this update is extremely small. 

As a side note: the tools library updates that introduced disco and NanoRange are still under review so the dependency is just pointing to a specific commit of the feature/python-views in tools. I will not merge this until we have tagged a tools release and point to that instead. That will just change the dependency files so we can already proceed with review.


timburculosis (approved):
🎉  looks great!  We'll work on the CMake updates for installations in the next MR.

PR #139, opened 2024-09-23
Fix/assignment

whaeck:
Replaces placement new in copy and move assignment to remove memory leaks.


joewpeterson (approved):
This update was manually checked for memory leaks using both leaks on local mac and valgrind on xdiv machine. No leaks were detected

